PR #5, opened 2016-05-31
fix for skimage deprecated import in acstools

SaraOgaz:
Reported by @dborncamp.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,4 +6,4 @@ ACSTOOLS
 .. |docs| image:: https://readthedocs.org/projects/acstools/badge/?version=latest
     :alt: Documentation Status
     :scale: 100%
-    :target: https://acstools.readthedocs.org/en/latest/?badge=latest
+    :target: https://acstools.readthedocs.io/en/latest/?badge=latest


dborncamp:
It would be good to note that this came in because the testing environment for this was UREKA SSBX which used scikit image version .10. Thank you for fixing that!

lib/acstools/satdet.py
```diff
@@ -7,6 +7,8 @@
     Only tested for ACS/WFC FLT and FLC images, but it should
     theoretically work for any instrument.
 
+    Requires skimage version 0.11.x or later
```
pllim: Nitpicking: End the sentence with a period.

lib/acstools/satdet.py
```diff
@@ -141,10 +144,11 @@
 from astropy.stats import biweight_midvariance, sigma_clipped_stats
 from astropy.utils.exceptions import AstropyUserWarning
 from scipy import stats
-from skimage import filter as filt
+from skimage import filters as filt
```
pllim: This import is not necessary anymore?

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = acstools
-version = 2.0.1
+version = 2.0.2
```
pllim: This should also be updated in `defsetup.py` for consistency although I don't think that file is used.

pllim:
p.s. It is "deprecated" :wink:

pllim:
p.p.s. If you want to practice squashing commits to increase your git-fu, you can use this as a practice to squash all your commits here into one. Not that it is necessary here, but just if you want some practice.

pllim:
Once the comments are addressed, I will merge this and make a release.

SaraOgaz:
Thanks @pllim!  I should be able to get to these comments on Friday.

SaraOgaz:
Closing this pull request, wasn't using a new branch *DOH*

PR #6, opened 2016-06-03
Fixed skimage.filters import, was deprecated

SaraOgaz:
Was using deprecated call.
Adding in comments from @pllim.

NOTE: Replaced #5.

### DIFF
--- a/lib/acstools/satdet.py
+++ b/lib/acstools/satdet.py
@@ -7,6 +7,8 @@ within an ACS/WFC image as published in
     Only tested for ACS/WFC FLT and FLC images, but it should
     theoretically work for any instrument.
 
+    Requires skimage version 0.11.x or later.
+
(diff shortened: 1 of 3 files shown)


lib/acstools/satdet.py
```diff
@@ -153,8 +156,8 @@
     warnings.warn('matplotlib not found, plotting is disabled',
                   AstropyUserWarning)
 
-__version__ = '0.3.1'
-__vdate__ = '25-Apr-2016'
+__version__ = '0.3.2'
+__vdate__ = '24-May-2016'
 __author__ = 'David Borncamp, Pey Lian Lim'
```
pllim: Sure you don't want to insert your name here? Changes look good otherwise.

pllim:
@SaraOgaz , @dborncamp -- A new release is done based on this fix at https://github.com/spacetelescope/acstools/releases/tag/2.0.2 . FYI.

Issue #14, opened 2016-09-08
acs_destripe cannot handle new subarrays


pllim:
For new subarray WFC1A-512 with the following FITS format:
```
Filename: test_blv_tmp.fits
No.    Name         Type      Cards   Dimensions   Format
0    PRIMARY     PrimaryHDU     235   ()              
1    SCI         ImageHDU        95   (2048, 512)   float32   
2    ERR         ImageHDU        48   (2048, 512)   float32   
3    DQ          ImageHDU        40   (2048, 512)   int16 
```

`StripeArray._get_ref_section()` is not extracting the correct DARK sub-section, resulting in an exception caused by array size mismatch when user attempts to run `acs_destripe_plus` on the subarray, as recommended.

Here are the computed variable values from that test image with the current code:
```python
refaxis1 = 4096
refaxis2 = 2048
sizaxis1 = 2048
sizaxis2 = 512
centera1 = 1037
centera2 = 1793
xdelta = 13
xlo = -1
xhi = 2047
ylo = 1536
yhi = 2048
```

pllim:
Fixed by #15.

PR #15, opened 2016-09-09
Fix acs_destripe to run in Python 3 and handle new subarrays

### LINKED ISSUE #13: acs_destripe is not Python 3 compatible
Need to change `xrange` to `range`.

pllim:
Fixes #13 and #14


pllim:
Merging now to kick of reg test on next run.

PR #19, opened 2016-09-24
**DO NOT MERGE** Apply correction factor for postflash

### LINKED ISSUE #17: acs_destripe to use new FLSHCORR algorithm
In `ingest_flash()`, there is an attempt to emulate FLSHCORR to the data. This needs to be updated to be in sync with spacetelescope/hstcal#16. While we are at it, probably cleaner to pluck out the CALACS emulation codes and stick them in a separate module. The fix should be merged (and released) at the same time as the mentioned HSTCAL pull request.

Note: Need to make a new release when merged. But how to coordinate with HSTCAL?

pllim:
**DO NOT MERGE. Pending approval from ACS Team.**

Fixes #17, which is equivalent to spacetelescope/hstcal#12.

Merge at the same time as spacetelescope/hstcal#16.

### DIFF
--- a/acstools/acs_destripe.py
+++ b/acstools/acs_destripe.py
@@ -62,8 +62,8 @@ from astropy.io import fits
 from .utils_calib import extract_dark, extract_flash, extract_flatfield
 
 __taskname__ = 'acs_destripe'
-__version__ = '0.8.2'
-__vdate__ = '22-Sep-2016'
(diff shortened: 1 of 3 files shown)


pllim:
See https://github.com/spacetelescope/hstcal/pull/16#issuecomment-278374699

PR #22, opened 2017-03-21
Update acstools to the latest bitmask in stsci.tools

mcara:
Update ``acstools`` to use latest ``bitmask`` functions - see https://github.com/spacetelescope/stsci.tools/pull/29

### DIFF
--- a/acstools/acs_destripe.py
+++ b/acstools/acs_destripe.py
@@ -551,7 +551,7 @@ def perform_correction(image, output, stat="pmode1", maxiter=15, sigrej=2.0,
 def _mergeUserMaskAndDQ(dq, mask, dqbits):
     from stsci.tools import bitmask  # Optional package dependency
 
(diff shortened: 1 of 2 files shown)


pllim:
This seems non-backward compatible. `stsci.tools` is an *optional* dependency here, so its version compatibility is not checked on install. You need to add a version check right after that local import, or make this backward-compatible somehow.

mcara:
@pllim take a look at the second attempt.

mcara:
Ah, I just saw your approval. Thanks!

pllim:
I don't see any need to do a immediate release of this change on AstroConda, so I'll wait and fold it into the next release whenever that may be. (Might have to do one before 2017.2 anyway but not guaranteed.)

pllim:
Wait, wait... Write a test! :stuck_out_tongue_closed_eyes:

Issue #26, opened 2017-06-30
Add gen2 CTE ISR to docs


jamienoss:
At the time of resolving #25 no new ISR was available. When this is available, it needs to be integrated into the docs.

jamienoss:
@JayVB Norman stated that you are authoring this? If you could keep me posted on when this is complete, that would be great, thanks.

jamienoss:
You mean the algorithm didn't change much from that used in calwf3, however, the change from the correction algorithm used in calacs is significant. Isn't this ISR for ACS or is it something different? I'm completely unfamiliar with the nature of an ISR.

JayVB:
I'm putting together an ISR (Instrument Science Report) on the pinning of the model in ACS/WFC and WFC3/UVIS... it's a joint report, but Norma and Elena may prefer to have it separated.  I'll talk to them in July.  Yes, the ACS algorithm is different from before, but it's not that different.... the fundamental idea is the same.

________________________________

PR #27, opened 2017-06-30
Update CALACS doc to include new CTE Gen 2 options

### LINKED ISSUE #25: Update CALACS doc to include new Gen 2 options
This needs to be updated:
http://acstools.readthedocs.io/en/latest/calacs_hstcal.html

Source is here:
https://github.com/spacetelescope/acstools/blob/master/doc/source/calacs_hstcal.rst

Also, did this change?
http://acstools.readthedocs.io/en/latest/calacs.html

jamienoss:
resolves #25

### DIFF
--- a/doc/source/calacs_hstcal.rst
+++ b/doc/source/calacs_hstcal.rst
@@ -57,12 +57,29 @@ CALACS supports several command line options:
   * Turn off parallel processing for PCTECORR. Try this option if you encounter
     problems running CALACS with PCTECORR=PERFORM.
 
+* --nthreads <N>
+
+  * Specify the number of threads used for PCTECORR.
+    The default is 'greedy' and will use that specified by the system environment variable OMP_NUM_THREADS.
+    If N > ``OMP_NUM_THREADS``, ``OMP_NUM_THREADS`` will be used instead. The option -1 takes precedence.
+
+* --ctegen <1|2>
+
+  * Specify which generation CTE correction to use, the default is 2. Gen 1 (officially deprecated) refers to
+    the correction algorithm used in calacs version pre 9.2.0 in relation to the following ISRs
+    ACS ISR 2010-03 and 2012-03. Gen 2 refers to the new CTE correction algorithm implemented in calacs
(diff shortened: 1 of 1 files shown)


jamienoss:
Once this has been merged and #26 resolved we should tag the next acstools until HSTCAL1.3.0 has been tagged.

doc/source/calacs_hstcal.rst
```diff
@@ -121,8 +138,8 @@ For post-SM4 full-frame WFC exposures, it also includes:
 Pixel-Based CTE Correction (PCTECORR)
 =====================================
 
-For all full-frame WFC exposures, pixel-based CTE correction (ACS ISR 2010-03
-and 2012-03) is applied in ACSCTE that occurs after the ACSCCD series;
```
pllim: We should keep the old ISRs in case user somehow desires Gen 1 correction. Just make it clear that they are for Gen 1.

doc/source/calacs_hstcal.rst
```diff
@@ -174,30 +202,17 @@ not recommended unless you know what you are doing.
     * 1 - Perform noise smoothing
     * 2 - No noise smoothing
 
-  * Overwrites NSEMODEL in PCTETAB.
-
-* PCTERNCL
```
pllim: Similar to my comment above, we should keep Gen 1 keywords around.

jamienoss:
I did think about whether I should keep both and split them out as you describe, however, I do know that Norman would prefer as little exposure and documentation of the old algorithm as possible. This is why I decided not to include excessive detail of the deprecated version.

To be honest, I doubt he even likes the idea of the option being documented at all.

Perhaps we should pass this one up the ladder?

I'm personally ok with keeping the old info as you would like, what do you think?

pllim:
Sure, Norman can decide. Thanks!

jamienoss:
Norman Grogin's email comment:

> This is fine by me.  I do not want to be encouraging people
to use the Gen1 version at all.  As you know, I would have
preferred that Gen1 capability be stripped out of the 
Gen2 PCTECORR entirely.
>
>The verbiage with the "--ctegen" flag is ample, imho.

PR #34, opened 2017-07-31
Compatibility with Astropy 2.0

pllim:
Replace FITS "clobber" keyword with "overwrite". The new keyword was introduced in Astropy 1.3. The old one deprecated in 2.0 and will be removed in 3.0. Do not confuse this with the local "clobber" keyword in "acs_destripe", as there is no plans to deprecate *that*.

Note for future: When our minimum Astropy requirement has gone beyond 1.3 (not sure who decides that), "clobber" part of this logic can be removed altogether.

### DIFF
--- a/acstools/acs_destripe.py
+++ b/acstools/acs_destripe.py
@@ -55,8 +55,10 @@ from __future__ import absolute_import, division, print_function
 import logging
 
 # THIRD-PARTY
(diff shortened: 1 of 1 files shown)


pllim:
Merging now so it goes into dev for testing.

PR #38, opened 2017-10-03
Pass arbitrary options to executables

### LINKED ISSUE #33: Add ability to pass arbitrary options through to called executable
This is effectively a backdoor to passing any and all options (whether they work or not) through to the underlying executable.

_c.f._ #31 

### Update:
Not applicable, at least under this issue, to ``destripe`` or ``destripe plus``. It covers the other wrappers in acstools:

* acs2d
* acsccd
* acscte
* acsrej
* acssum
* calacs

jamienoss:
Resolves #33

### DIFF
--- a/acstools/acs2d.py
+++ b/acstools/acs2d.py
@@ -32,7 +32,7 @@ __vdate__ = "10-Oct-2014"
 __all__ = ['acs2d']
 
 
-def acs2d(input, exec_path='', time_stamps=False, verbose=False, quiet=False):
+def acs2d(input, exec_path='', time_stamps=False, verbose=False, quiet=False, exe_args=None):
     """
     Run the acs2d.e executable as from the shell.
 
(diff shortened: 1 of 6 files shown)


jamienoss:
@tddesjardins hang tight.

jamienoss:
@pllim thanks for the approval.

> Documentation wise, "arbitrary arguments" is not very informative to users; but that should be sufficient for now.

Do you have anything else in mind? I don't really see how it's _not_  informative. The logic is that "arbitrary" clearly implies that no checking is conducted to whether the passed arguments are valid. I.e. anything can be passed through.

pllim:
As a user, I wouldn't know what to pass in as "arbitrary". It would be better to explicitly document all the acceptable options, but that can be battle for another day.

jamienoss:
> As a user, I wouldn't know what to pass in as "arbitrary". It would be better to explicitly document all the acceptable options, but that can be battle for another day.

That would defeat the entire purpose and be poor design. That being said perhaps a link should be given to the actual documentation for the C executables. We really don't want to be duplicating anything.

jamienoss:
>>As a user, I wouldn't know what to pass in as "arbitrary". It would be better to explicitly document all the acceptable options, but that can be battle for another day.
>
>That would defeat the entire purpose and be poor design. That being said perhaps a link should be given to the actual documentation for the C executables. We really don't want to be duplicating anything.

Actually... since this passes arbitrary arguments I'll just add to the docs ``--help'`` to get the full usage.

pllim:
Yes, mentioning `--help` would be helpful. Thanks.

jamienoss:
Going to merge and defer to #39

PR #43, opened 2017-11-03
Added function to check for weird custom subarrays

pllim:
This should more or less sniff out the custom subarrays from the past that cannot get through the pipeline because they have overscan but lack an entry in the their OSCNTAB.

### DIFF
--- a/acstools/utils_calib.py
+++ b/acstools/utils_calib.py
@@ -8,11 +8,12 @@ import warnings
 # THIRD-PARTY
 import numpy as np
 from astropy.io import fits
+from astropy.table import Table
 from astropy.utils.exceptions import AstropyUserWarning
 
 __all__ = ['extract_dark', 'extract_flash', 'extract_flatfield',
            'from_irafpath', 'extract_ref', 'find_line', 'get_corner', 'get_lt',
-           'from_lt']
+           'from_lt', 'is_weird_subarray']
 
 
 def extract_dark(prihdr, scihdu):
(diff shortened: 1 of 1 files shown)


acstools/utils_calib.py
```diff
@@ -464,3 +465,81 @@ def _nint(x):
...
+    j8dm03fxq_raw.fits True
+    j8eua7qqq_raw.fits False
+
+    """
+    is_weird = False
+
+    # Dimension of fullframe dark on each chip.
+    max_y = 2048
+    max_x = max_y * 2
+
```
mdlpstsci: If the true maximum *full* fullframe values can be obtained from the FITS header, it would be better to do this rather than hardcode dimensions.

acstools/utils_calib.py
```diff
@@ -464,3 +465,81 @@ def _nint(x):
...
+        ny = scihdr['NAXIS2']
+        darkfile = prihdr['DARKFILE'].replace('jref$', os.environ['jref'])
+
+        with fits.open(darkfile) as pfdrk:
+            same_size, rx, ry, x0, y0 = find_line(pf[1], pfdrk[1])
+
+        x1 = x0 + nx
+        y1 = y0 + ny
+
+        # Assume rx and ry are always 1.
```
mdlpstsci: Add better comments here so it is clear that images which apparently do not fulfill the if statement criteria (as I recall from Tyler's testing) will (or are expected to) process correctly with CALACS.

pllim: This is just a one-off tool. If we want to document why `rx` and `ry` are always 1, they should be done both here and in HSTCAL. I am also not sure of the best wording, so I'll have to dig around. Thanks for pointing this out.

pllim:
This turns out is not what @tddesjardins needs for his reporting to DMS. He'll submit his own solution as a separate PR.

PR #45, opened 2017-11-08
Added functionality to check for custom subarrays

tddesjardins:
These additions in utils_calib.py seem to catch all custom subarrays and also returns information about the number of bias columns.

### DIFF
--- a/acstools/utils_calib.py
+++ b/acstools/utils_calib.py
@@ -464,3 +464,128 @@ def _nint(x):
     else:
         y = x - 0.5
     return int(y)
+
+
+def get_hdr_vals(root):
+
+    """Retrieve header keyword values from RAW and SPT
+    FITS files to pass on to check_oscntab() and
+    check_overscan().
+
+    Parameters
+    ----------
+    root: str
+        Rootname of the observation. Can be relative path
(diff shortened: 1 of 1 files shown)


pllim:
Thank you for working on this, @tddesjardins ! I'll have a look soon. 😄

pllim:
From looking at the hardcoded values, it seems like you only checked WFC subarrays. Are there no custom HRC subarrays? This OSCNTAB also affects HRC, so just want to double check.

tddesjardins:
I'm not sure if there are custom HRC subarrays off the top of my head, but will look into this now.

pllim:
Given that HRC is only 1k by 1k, it is probably unlikely users would be compelled to make their own subarrays, but with the coronograph, I can't be 100% sure. Thanks for checking!

tddesjardins:
Yes, in fact the old instrument handbook says they find the likelihood a user would want to do a custom subarray with HRC to be remote, but says they can do it for nearly arbitrary image sizes (what a headache!).

pllim:
@tddesjardins , hopefully my edits over at tddesjardins/acstools#1 would make this check easier.

PR #51, opened 2018-01-11
ACSREJ to accept readnoise_only instead of newbias

### LINKED ISSUE #50: Update ACSREJ wrapper for 2018.1 delivery
Deadline: Jan 15, 2018

Modify Python wrapper in conjuction with spacetelescope/hstcal#288.

c/c @mdlpstsci

pllim:
Fix #50 . This is in conjunction with spacetelescope/hstcal#288 .

### DIFF
--- a/acstools/acsrej.py
+++ b/acstools/acsrej.py
@@ -28,6 +28,7 @@ For help usage use ``exe_args=['--help']``
 # STDLIB
 import os
 import subprocess
+import warnings
 
 __taskname__ = "acsrej"
 __version__ = "2.0"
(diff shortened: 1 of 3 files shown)


acstools/acsrej.py
```diff
@@ -215,6 +231,4 @@ def run(configobj=None):
            crradius=configobj['crradius'],
            crthresh=configobj['crthresh'],
            badinpdq=configobj['badinpdq'],
-           newbias=configobj['newbias'],
-           exe_args=configobj['exe_args']
```
pllim: Note: `exe_args` is not a part of the TEAL interface. In fact, TEAL interface does not understand Python list.

If we want to support arbitrary `exe_args` in TEAL, it needs to be read in as one long comma-separated string and then parsed into a list. Such work is beyond the scope of this PR. Also argument could be made that such an option can be very confusing for the user on the GUI side.

pllim:
As discussed offline, this should not be merged until spacetelescope/hstcal#288 is merged upstream.

PR #53, opened 2018-04-11
[WIP] Added pixel area map code

tddesjardins:
The pixel area maps (PAMs) needed to convert distorted FLT/FLC files to the undistorted frame have not been updated since 2004. Because of changes to the distortion model over time, it is now preferred to have users create PAMs based on the FLT/FLC file rather than provide one-size-fits-all files for download. 

This new module is a Python translation of the IRAF code created by Richard Hook in 2004 to create pixel area maps using a call to AstroDrizzle. It works for all three ACS detectors and I've tested the PAMs it creates against the old files, which only show minor differences of ~0.1% and are almost certainly due to updates to the distortion model since the old files were created in 2004.


tddesjardins:
As per @pllim 's request, we're requesting a release after the PR is merged.

pllim:
As per off-line discussion, I'll mark this as WIP (work in progress). Please let me know when this is ready. Thank you!

tddesjardins:
@mcara wrote a better PAM code and made it part of `stsci.skypac`, so I'm withdrawing this code from `acstools`.

mcara:
FYI, PR with PAM code: https://github.com/spacetelescope/stsci.skypac/pull/28

PR #58, opened 2018-05-31
Clarify acs_destripe_plus doc and clean up code

### LINKED ISSUE #56: acs_destripe_plus supports fullframe but doc is misleading
Turns out `acs_destripe_plus` always has logic to process fullframe but both inline comments and ACS documentation are misleading in stating that it is only for certain supported *subarrays*.

c/c @tddesjardins

### LINKED ISSUE #57: acs_destripe_plus does not support special PIXVALUE extension
`acs_destripe_plus` crash when image has an empty extension but `PIXVALUE` header keyword to indicate it is actually a constant-value image extension. This is a custom HST format widely used in HST data. The simplest solution is to use `stpyfits` module in `stsci.tools`.

c/c @tddesjardins

pllim:
- [x] Fix #56 
- [x] Fix #57

### DIFF
--- a/acstools/acs_destripe.py
+++ b/acstools/acs_destripe.py
@@ -57,8 +57,14 @@ import logging
 # THIRD-PARTY
 import astropy
 import numpy as np
-from astropy.io import fits
 from astropy.utils.introspection import minversion
+
+try:
+    # This supports PIXVALUE
+    from stsci.tools import stpyfits as fits
+except ImportError:
+    # Falls back to Astropy
+    from astropy.io import fits
 
 # LOCAL
 from .utils_calib import extract_dark, extract_flash, extract_flatfield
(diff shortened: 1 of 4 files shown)


pllim:
@tddesjardins , do you want to take this one for a spin? Does it work with the `PIXVALUE` images as you expected now?

tddesjardins:
To test this I did the following:

1. Retrieved an observation using the WFC1B-2K aperture (dataset JD5702JUQ).
2. Set BIASCORR, SINKCORR, and DQICORR to 'OMIT'  in the raw file and ran the image through `acsccd` to check the blv_tmp.fits file. Verified that the DQ array was stored using the PIXVALUE keyword in the header.
3. Deleted the blv_tmp.fits file, and then ran the raw file through `acs_destripe_plus` with the same calibration flags. The code completed normally and produced an flt.fits file.

This seems to behave as expected and I believe can be merged.

PR #60, opened 2018-07-26
Add acscte_forward_model functionality

jamienoss:
acstools functionality parity with https://github.com/spacetelescope/hstcal/pull/313

### DIFF
--- a/acstools/acscte.py
+++ b/acstools/acscte.py
@@ -1,6 +1,7 @@
 """
-The acscte module contains a function `acscte` that calls the ACSCTE executable.
-Use this function to facilitate batch runs of ACSCTE, or for the TEAL interface.
+The acscte module contains the functions `acscte` and `acscte_forward_model` that call the ACSCTE
+and ACSCTEFORWARDMODEL executables respectively.
+Use the `acscte` function to facilitate batch runs or for the TEAL interface.
 
 Only WFC full-frame and some 2K subarrays are currently supported. See
 `ACS Data Handbook <http://www.stsci.edu/hst/acs/documents/handbooks/currentDHB/>`_
@@ -38,6 +39,54 @@ __taskname__ = "acscte"
 __version__ = "1.0"
 __vdate__ = "13-Aug-2013"
 __all__ = ['acscte']
+
+def acscte_forward_model(input, exec_path='', time_stamps=False, verbose=False, quiet=False,
+           single_core=False, exe_args=None):
+    """
+    Run the acscteforwardmodel.e executable as from the shell.
+
+    Expect input to be ``*_blv_tmp.fits``.
+    Output is automatically named ``*_blc_tmp.fits``.
+
+    Parameters
+    ----------
+    input : str or list of str
+        Input filenames in one of these formats:
+
+            * a single filename ('j1234567q_blv_tmp.fits')
+            * a Python list of filenames
+            * a partial filename with wildcards ('\*blv_tmp.fits')
+            * filename of an ASN table ('j12345670_asn.fits')
+            * an at-file (``@input``)
+
+    exec_path : str, optional
+        The complete path to ACSCTEFORWARDMODEL executable.
+        If not given, run ACSCTEFORWARDMODEL given by 'acscteforwardmodel.e'.
+
+    time_stamps : bool, optional
+        Set to True to turn on the printing of time stamps.
+
+    verbose : bool, optional
+        Set to True for verbose output.
+
+    quiet : bool, optional
+        Set to True for quiet output.
+
+    single_core : bool, optional
+        CTE simulation in ACSCTEFORWARDMODEL will by default try to use all available
+        CPUs on your computer. Set this to True to force the use of just
+        one CPU.
+
+    exe_args : list, optional
+        Arbitrary arguments passed to underlying executable call.
+        Note: Implementation uses subprocess.call and whitespace is not
+        permitted. E.g. use exe_args=['--nThreads', '1']
+
+    """
+
+    return _acscte_base(input, exec_path=exec_path, time_stamps=time_stamps,
+                        verbose=verbose, quiet=quiet, single_core=single_core,
+                        exe_args=exe_args, exe='acscteforwardmodel.e')
 
 
 def acscte(input, exec_path='', time_stamps=False, verbose=False, quiet=False,
@@ -83,6 +132,15 @@ def acscte(input, exec_path='', time_stamps=False, verbose=False, quiet=False,
         permitted. E.g. use exe_args=['--nThreads', '1']
 
     """
+
+    return _acscte_base(input, exec_path=exec_path, time_stamps=time_stamps,
+                        verbose=verbose, quiet=quiet, single_core=single_core,
+                        exe_args=exe_args, exe='acscte.e')
+
+
+def _acscte_base(input, exec_path='', time_stamps=False, verbose=False, quiet=False,
+           single_core=False, exe_args=None, exe='acscte.e'):
+
     from stsci.tools import parseinput  # Optional package dependency
 
     if exec_path:
@@ -90,7 +148,7 @@ def acscte(input, exec_path='', time_stamps=False, verbose=False, quiet=False,
             raise OSError('Executable not found: ' + exec_path)
         call_list = [exec_path]
     else:
-        call_list = ['acscte.e']
+        call_list = [exe]
 
     # Parse input to get list of filenames to process.
     # acscte.e only takes 'file1,file2,...'


acstools/acscte.py
```diff
@@ -1,6 +1,7 @@
 """
-The acscte module contains a function `acscte` that calls the ACSCTE executable.
-Use this function to facilitate batch runs of ACSCTE, or for the TEAL interface.
+The acscte module contains the functions `acscte` and `acscte_forward_model` that call the ACSCTE
+and ACSCTEFORWARDMODEL executables respectively.
+Use the `acscte` function to facilitate batch runs or for the TEAL interface.
```
pllim: Please clarify whether the batch run is for both or only ACSCTE.

Please add an example of its usage in Python a few lines below.

acstools/acscte.py
```diff
@@ -39,6 +40,54 @@
 __vdate__ = "13-Aug-2013"
 __all__ = ['acscte']
```
pllim: Please add `'acscte_forward_model'` inside `__all__` list.

acstools/acscte.py
```diff
@@ -39,6 +40,54 @@
...
 
+def acscte_forward_model(input, exec_path='', time_stamps=False, verbose=False, quiet=False,
+           single_core=False, exe_args=None):
+    """
+    Run the acscteforwardmodel.e executable as from the shell.
+
+    Expect input to be ``*_blv_tmp.fits``.
+    Output is automatically named ``*_blc_tmp.fits``.
```
pllim: I am not sure about this one. I thought it is theoretically the other way around? Maybe @jryon can double check. I wasn't involved in the API discussion.

acstools/acscte.py
```diff
@@ -39,6 +40,54 @@
 __vdate__ = "13-Aug-2013"
 __all__ = ['acscte']
 
+def acscte_forward_model(input, exec_path='', time_stamps=False, verbose=False, quiet=False,
+           single_core=False, exe_args=None):
```
pllim: Nitpick: Indentation doesn't match.

acstools/acscte.py
jryon: The output is the input with the appended suffix "_ctefmod.fits". I think the input filename doesn't technically matter, but will likely be files like FLCs and BLC_TMPs.

jamienoss: Sorry, I didn't pay attention to this and just copied what was there before. TBH I wouldn't merge this if it were up to me. The bulk principle is what I was aiming for, i.e. no code dupe. The small bits here and there probably need more attention.  I left this PR open to co-edits so anyone can go for it if they like. Or copy/pull it etc. I don't really have time but, like I said, this is the sort of thing this change should look like (without rewriting it all and OOing with classes).

acstools/acscte.py
jamienoss: >Please clarify whether the batch run is for both or only ACSCTE.

I thought "Use _this_ function" -> "Use the `acscte` function" was clarification, no?

>Please add an example of its usage in Python a few lines below.

 Anyone...? 😉

jamienoss:
> As mentioned briefly in #61, this implementation here does not support a separate TEAL interface for the forward modeler. For TEAL to work, forward modeler needs to be in its own module with its own TEAL parameters.

I didn't think we still supported this? But then I don't really know what it is either.

If we still really require/support it, that's a shame, as I think it makes much more sense for there to exist a CTE module that contains all the CTE functionality, i.e. both ``acscte`` & ``acscte_forward_model`` (which, BTW, I now wish we called ``acscteSim``). I mean that's the very design use of modulization, right, to contain semantically-related code.

If we're on the shelf regarding TEAL support, I would dump it in favour of having actual modules.

The design of this PR can still persist though, the new mod just needs to import the base func and contain the new wrapper + TEAL stuff.

mdlpstsci:
I was not sure where to put this comment so I will reference everything in the interests of coverage (IT #323, PR #61). I do not want the upcoming build due date (we should close our changes/testing by 07 August 2018) for the DMS pipeline to influence the proper fix and associated testing for this issue.  While Jenna's workaround sounds like it is well tested, my concern about releasing it as part of the ACSTOOLS package in the upcoming release is its (probable) incompatibility with the true fix which will be done in the C code, maintaining a thin Python wrapper.  This fix may well come on the heels of the big pipeline release so it could get out to the public in a timely fashion.  I want to avoid confusion for the community with having different versions of this new ACSTOOL which are incompatible with one another and need different underlying HSTCAL software.  I also think that making the community wait just a little bit longer for robustly implemented and tested functionality is a good thing.

Warren and I have discussed this, and our work flow is such that we can release utilities out-of-step (if need be) releases of desired functionality which some might consider experimental.  In some ways this type of release might be best as this will give the community the opportunity to "play" with this functionality and to give us feedback with how they are using it.  Of course, functionality like this will be an official part of the next pipeline release which could be a quick fix (2018.3b or the next build 2019.1).  I suggest we have a faces-to-faces discussion to clarify any issues/complaints/questions. @jamienoss @jryon @pllim @stsci-hack @mdlpstsci.

stscicrawford:
I agree with @mdlpstsci here that if a satisfactory solution can not be implemented well before the freeze date, then the wrapper functionality should not be included but added as a quickfix afterwards. 

Nonetheless, it would seem that a workable solution fo ACS has been identified (updating how [nimsets is calculated](https://github.com/spacetelescope/hstcal/blob/2.0.0/pkg/acs/calacs/lib/getacskeys.c#L94-L95) -- and I think @jamienoss has made multiple reasonable suggestions for how that might be fixed), so I would suggest trying to implement a workable solution before next week.  We have test data sets and in case it does have any other impact, it could be picked up during the testing of the build.  

If a workable solution is then implemented, then I'd suggest updating #61 to only be a light wrapper as it already supplied by @jryon and is working and tested by the user. 

If further discussion is needed, then let's have a short meeting as soon as possible.

jamienoss:
I think one of the best workarounds might also be the most reusable - that is to add an entirely new function to acstools, or maybe stak/astropy, that can strip a fits file to just its ``SCI``, ``ERR``, & ``DQ`` imsets. I.e. it ingests a multi-extension fits file, computes the number of imsets and then produces a new file with just these, it strips out the unwanted extensions. I would imagine it to also take an optional nucleus extension, one that defines the imset, i.e. ``setContainsEXTNAME``, e.g. ``setContainsEXTNAME="SCI"``.

mdlpstsci:
The rough plan suggested by @stscicrawford sounds fine, but I have no feeling as to who is doing what and when they are doing it, and if something is expected of me(!).  I personally am tightening up some CALACS C code while waiting for the ACS team to give me the OK on the science aspects of a requested update.  @pllim is waiting in the proverbial wings in case we would like to her to make some Python changes in support of this issue, @jamienoss has put forth some viable suggestions, and @jryon has offered her script.  So I have to ask, @jamienoss, I would like to confirm that you are working this issue as I know you have other commitments.  If my assumption is true, then are you working and C and Python aspects, or will @pllim need to be involved?  I am just trying to get a clearer picture on what is happening, and who needs to step-up.  I am not making any assumptions about this being done in time for the due date for the code though I have concern if @pllim will be pressed into service at the last moment.  I am just trying to feel my way in the dark.

jamienoss:
@mdlpstsci I'm (intermittently) working on the fix to go into HSTCAL, it's just requiring me to hit CFITSIO directly, under the hood so to speak, and 1) I've never touched it before, & 2) the documentation is poor, 3) so is its design - global status etc (FYI looks like hstcal/hstio does this because cfitsio does, it's the root).

I wouldn't hold your breath for me though, I might be able to get it working in the next day or two (maybe). If you want just strip the fix out of @jryon PR and merge that. Perhaps adding some comment to the doc that it takes files with only imsets present and maybe giving a workaround for this?

Either way, I don't remember ``acstools`` being a deal-breaker for OPs to get on with their testing etc, or even for us and our release.

So there are only 2 outstanding things:

 * ``ACSTOOLS`` wrapper for ``acscteforwardmodel.e`` #60 | #61 
 * An actual fix for  https://github.com/spacetelescope/hstcal/issues/323

philhodge:
@jamienoss wrote "3) so is its design - global status etc (FYI looks like hstcal/hstio does this because cfitsio does, it's the root)"
hstcal/hstio was written by Allen Farris in 1995.  The first release of cfitsio was in 1996, but in any case, I doubt that Allen would have been influenced by the design of cfitsio.  Prior to cfitsio there was fitsio, which was written in Fortran; status was an argument in the calling sequence, but it was updated in-place rather than being the function value.
In what sense does cfitsio use global status?  In most cfitsio functions, status is an input argument, and the function returns status, after possibly modifying the value.

jamienoss:
@philhodge you're right I was getting confused with something else.

pllim:
@stscicrawford , @mdlpstsci , et al. -- Whatever that ACS Team can live with is fine by me. They are the main stakeholder for this package.

pllim:
Superseded by #61

PR #61, opened 2018-07-26
Add wrapper for acscteforwardmodel.e

jryon:
Includes a workaround for input files with >6 extensions.


acstools/acscteforwardmodel.py
```diff
@@ -0,0 +1,254 @@
...
+    if exe_args:
+        call_list.extend(exe_args)
+
+    subprocess.check_call(call_list)
+
+    # Restore extra FLC extensions to forward model
+    restore_files(infiles)
+
+
+def make_tmp_files(infiles):
```
jamienoss: This 'workaround' belongs in the underlying exec. With it here it expands the wrapper and really these particular ones should be "wafer thin" (add french accent). Also, there's a chance that it could break or complicate functionality in the underlying exec in the future by adding this functional dependence in the wrong place.

jamienoss: p.s. I would also make this "hidden" and made the name more explicit to reduce potential namespace clashes, e.g. ``_create_ctefm_tmp_input_files()``.

Though I would have also made it more semantically descriptive, e.g. ``_create_max_n_imset_tmp_file`` (or something). If you really want just six, 1st write the generic func and then wrap it with the ``n`` made explicit.

acstools/acscteforwardmodel.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    If more than 6 extensions, take SCI, ERR, and DQ extensions and make a
```
jamienoss: Is this even what is needed? I didn't think the issue was specific to 6, isn't the 6 just an artifact of the input file that was used? That after the 1st 3 imsets (6 exts) that file contained other extensions that were not imsets. I think what you actually want is to iterate over the ``hdu.info`` tab and just strip out the full imsets. This is at least what I'm thinking about for the fix in the C code.

jamienoss:
Workaround for https://github.com/spacetelescope/hstcal/issues/323

acstools/acscteforwardmodel.py
jryon: Yes, you are right that the issue isn't specific to 6. This was simply the first method I thought of that would separate both full-frame and subarray images with the distortion extensions from those without them. Identifying the imsets is probably a better way to go.

pllim:
FWIW practicality beats purity here. I actually prefer the forward modeler to be its own module and has its own TEAL interface. Code duplication is not a deal breaker for me; cleaning that up should be a separate issue as duplication is a package-wide problem. And it is obviously well tested as @jryon already uses it for her own work. I am actually in favor of closing #60 and choose this one.

jamienoss:
>FWIW practicality beats purity here

For something as simple as this, that point is totally defunct. I.e. it is trivial to make this degree of practicality also pure(r).

>Code duplication is not a deal breaker for me; cleaning that up should be a separate issue as duplication is a package-wide problem.

There's a very strong incentive to NOT make it worse in the interim though. Arguing not to ensure that each addition is not cleaner than before on the grounds of "the before" being ugly is not valid argument. It only trends to making things worse rather than at least holting it.

>I am actually in favor of closing #60 and choose this one.

I would advise not merging this as is, the fix for https://github.com/spacetelescope/hstcal/issues/323 shoudl not exist in this package.

acstools/acscteforwardmodel.py
jamienoss: A shorter workaround would/could be to just copy the entire file to a tmp and then change ``NEXTEND`` (in the new tmp file) to the number of imsets. This would decrease the size of the wrapper.

```python
from astropy.io import fits
hduList = fits.open('<fname>')
nimsets = [hdu.name for hdu in hduList].count('SCI')
```  

However, I would argue that the briefness of this workaround leads to not being included at all, i.e. it remains a workaround and not fix (i.e. it doesn't get merged).

jamienoss: I am making an assumption here about the definition of an imset that I am unsure is a vlid one. _c.f._ my question in https://github.com/spacetelescope/hstcal/issues/323#issuecomment-408457258

jamienoss: Actually, you'd need ``NEXTEND = nimsets*3`` due to [this ugly line of code](https://github.com/spacetelescope/hstcal/blob/2.0.0/pkg/acs/calacs/lib/getacskeys.c#L95):
```c
...
/* Three extensions per SingleGroup. */
# define EXT_PER_GROUP 3
...
/* Convert number of extensions to number of SingleGroups. */
acs->nimsets = nextend / EXT_PER_GROUP;
...
```

jamienoss: _c.f._ https://github.com/spacetelescope/hstcal/issues/323#issuecomment-408524951
```python
nimsets = len(set([hdus.ver for hdus in hduList]))
```

mdlpstsci:
@jryon HSTCAL https://github.com/spacetelescope/hstcal/pull/330 resolves IT https://github.com/spacetelescope/hstcal/issues/323 in a satisfactory manner.  Following @stscicrawford plan seen in PR #60, we can proceed to use your wrapper at this time after it is modified to be a thin wrapper since the HSTCAL software has been updated.  Do you have time to modify your wrapper and test?

pllim:
`acstools/._acscteforwardmodel.py` binary file is perhaps added accidentally? Please back it out and use `.gitignore` to ignore it.

PR #63, opened 2018-08-07
Update RELIC implementation

jhunkeler:


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "relic"]
 	path = relic
-	url = https://github.com/jhunkeler/relic.git
+	url = https://github.com/spacetelescope/relic.git
```
pllim: Hah!

jhunkeler: Yeah it happened after several years :grimacing:

PR #74, opened 2018-10-02
EXP: Check to see if ci-watson PR breaks anything (DO NOT MERGE)

pllim:
**DO NOT MERGE**

Just to check to see if spacetelescope/ci_watson#16 broke anything over here.


pllim:
Looks good! But will keep this open until `ci-watson` PR is merged, just in case we need to retest.

PR #78, opened 2019-01-23
Zeropoint API

### LINKED ISSUE #77: Enhancement: Zeropoint Calculator API
The [ACS Zeropoints Calculator]( https://acszeropoints.stsci.edu/) is a webtool developed by the ACS Team to provide users with an interface to compute time-dependent zeropoints without having to download the required component files from CRDS. Users submit a `date` and `detector` which together comprise an "obsmode" used by `pysynphot` to compute the `STmag`, `VEGAmag`, `ABmag` zeropoints for all the ACS filters (or just a single one).

We have had several users ask (e.g. INC0130965) if there is anyway to script the process of computing zeropoints for a large number of date and detector combinations. I have re-written the Zeropoints Calculator to work using `GET` methods, rather than `POST`. Hence, the user supplied information is now passed via a query string embedded in the URL which means we can finally allow users to batch process things. Next, to give users tools to utilize these updates I wrote a new module in `acstools` called `acszpts` that allows users to programmatically query the Zeropoints Calculator and returns the results as an `astropy.table.Table` with units. The exact API is still a work in progress and I am open to any suggestions. I am opening this issue to keep track of things and provide a forum for discussion.

Example Usage:
```python
In [1]: from acstools import acszpt
The following tasks in the acstools package can be run with TEAL:
          acs2d                  acs_destripe           acs_destripe_plus     
          acsccd                    acscte              acscteforwardmodel    
          acsrej                    acssum                    calacs
PixCteCorr is no longer supported. Please use acscte.

In [2]: zpt_query = acszpt.Query(date='2016-04-04',detector='WFC')

In [3]: zpt_query.submit_request()

In [4]: zpt_query.parse_response()

In [5]: zpt_table = zpt_query.format_results()

In [6]: zpt_table
Out[6]: 
<Table length=13>
FILTER PHOTPLAM        PHOTFLAM         STmag  VEGAmag  ABmag 
       Angstrom erg / (Angstrom cm2 s) mag(ST)   mag   mag(AB)
 str6  float64         float64         float64 float64 float64
------ -------- ---------------------- ------- ------- -------
 F435W   4329.2              3.148e-19  25.155  25.763  25.665
 F475W   4746.2              1.827e-19  25.746  26.149  26.056
 F502N   5023.0              5.259e-18  22.098  22.365  22.285
 F550M   5581.5               3.99e-19  24.898  24.825  24.856
 F555W   5360.9              1.963e-19  25.667  25.713  25.713
 F606W   5922.0              7.811e-20  26.668  26.405  26.498
 F625W   6312.0              1.188e-19  26.213  25.735  25.904
 F658N   6584.0               1.97e-18  23.164  22.381  22.763
 F660N   6599.4              5.156e-18  22.119  21.428  21.714
 F775W   7693.2              9.954e-20  26.405  25.272  25.667
 F814W   8045.0              7.046e-20   26.78  25.517  25.944
F850LP   9033.2               1.52e-19  25.945  24.332  24.858
 F892N   8914.8              1.502e-18  23.458  21.905    22.4
```

nmiles2718:
I have added a new module for `acstools` called `acszpt`, which address the need for an interface to programmatically query the [ACS Zeropoints Calculator](https://acszeropoints.stsci.edu/). The module is designed to only work with valid combinations of the ACS instruments, their filters, and observation dates. If the users query fails to meet the proper criteria, an error is raised along with messages for each input that failed.  

closes #77

### DIFF
--- a/acstools/acszpt.py
+++ b/acstools/acszpt.py
@@ -86,7 +86,14 @@ class Query(object):
         One of the three channels on ACS: HRC, SBC, or WFC
     filt : str, optional
         One of valid filters for the chosen detector
-            * HRC and WFC:
+            * HRC:
+                 F220W, F250W, F330W,
+                 F344N, F435W, F475W,
+                 F502N, F550M, F555W,
+                 F606W, F625W, F658N, F660N,
+                 F775W, F814W, F850LP, F892N
+
+            * WFC:
                  F435W, F475W,
                  F502N, F550M, F555W,
                  F606W, F625W, F658N, F660N,
@@ -98,21 +105,13 @@ class Query(object):
     Attributes
     ----------
     date : str
-        Input date in the following ISO format, YYYY-MM-DD
+        **Attribute** for storing the user specified date.
     detector : str
-        One of the three channels on ACS: HRC, SBC, or WFC
-    filt : str, optional
-        One of valid filters for the chosen detector
-    zpt_table : astropy.table.Table object
-        Results returned from the ACS Zeropoint Calculator
-
-    Methods
-    -------
-    fetch() :
-        Submits the request to compute the zeropoints and returns an Astropy
-        Table
-
-
+        **Attribute** for storing the user specified detector.
+    filt : str, None
+        **Attribute** for storing the user specified filter (if supplied).
+    zpt_table : astropy.table.Table
+        **Attribute** for storing the results returned by the ACS Zeropoint Calculator
     """
 
     def __init__(self, date, detector, filt=None):
@@ -296,16 +295,18 @@ class Query(object):
         # Loop through each column and set the appropriate units
         for i, col in enumerate(tab.colnames):
             tab[col].unit = data_units[i]
-
         self.zpt_table = tab
 
     def fetch(self):
-        """ Submit the request and format the results
+        """ This method is used to submit a request to the ACS Zeropoints Calculator
+
+        Submit the request and format the results into an Astropy Table object
+        with the correspond units applied to each column
 
         Returns
         -------
-        If the request is successful, returns an Astropy table with the correct
-        units. If the request fails, returns None.
+        astropy.table.Table or None
+            Returns a table if the request was successful and None otherwise
         """
         valid_inputs = self._check_inputs()
         if valid_inputs:
--- a/acstools/acszpt.py
+++ b/acstools/acszpt.py
@@ -12,8 +12,8 @@ combination. In the first case, the query
 will return the zeropoint information for the specific filter and detector at
 specified date. In the second case, the query will return the zeropoint
 information for all the filters for the desired detector at the specified date.
-In either case, the result will be an Astropy table, where each column has the
-correct units attached.
+In either case, the result will be an ``astropy.table.QTable`` where each column
+is an ``astropy.units.quantity.Quantity`` object with the appropriate units attached.
 
 Examples
 --------
@@ -57,6 +57,25 @@ FILTER PHOTPLAM        PHOTFLAM         STmag  VEGAmag  ABmag
 ------ -------- ---------------------- ------- ------- -------
  F435W   4329.2              3.148e-19  25.155  25.763  25.665
 
+
+Retrieve the zeropoint information for the F435W filter for WFC at multiple dates:
+
+>>> from acstools import acszpt
+>>> dates = ['2004-10-13', '2011-04-01', '2014-01-17', '2018-05-23']
+>>> queries = []
+>>> for date in dates:
+...     q = acszpt.Query(date=date, detector='WFC', filt='F435W')
+...     zpt_table = q.fetch()
+...     # Each object has a zpt_table attribute, so we save the instance
+...     queries.append(q)
+>>> for q in queries:
+...     print(q.date, q.zpt_table['PHOTFLAM'][0], q.zpt_table['STmag'][0])
+2004-10-13 3.074e-19 erg / (Angstrom cm2 s) 25.181 mag(ST)
+2011-04-01 3.138e-19 erg / (Angstrom cm2 s) 25.158 mag(ST)
+2014-01-17 3.144e-19 erg / (Angstrom cm2 s) 25.156 mag(ST)
+2018-05-23 3.152e-19 erg / (Angstrom cm2 s) 25.154 mag(ST)
+>>> type(queries[0].zpt_table['PHOTFLAM'])
+astropy.units.quantity.Quantity
 """
 from six.moves.urllib.request import urlopen
 from six.moves.urllib.error import URLError
@@ -66,8 +85,9 @@ import logging
 from collections import OrderedDict
 
 import astropy.units as u
-from astropy.table import Table
+from astropy.table import QTable
 from bs4 import BeautifulSoup
+import numpy as np
 
 __taskname__ = "acszpt"
 __author__ = "Nathan Miles"
@@ -146,13 +166,24 @@ class Query(object):
                          '&{1}_filter={2}'.format(self.date,
                                                   self.detector,
                                                   self.filt))
-
-        self._zpts = OrderedDict()
+        # ACS Launch Date
         self._acs_installation_date = dt.datetime(2002, 3, 7)
+        # The farthest date in future that the component and throughput files
+        # are valid for. If input date is larger, extrapolation is not valid.
+        self._extrapolation_date = dt.datetime(2021, 12, 31)
         self._msg_div = '-' * 79
         self._valid_detectors = ['HRC', 'SBC', 'WFC']
         self._response = None
         self._failed = False
+        self._data_units = {
+            'FILTER': u.dimensionless_unscaled,
+            'PHOTPLAM': u.angstrom,
+            'PHOTFLAM': u.erg / u.cm ** 2 / u.second / u.angstrom,
+            'STmag': u.STmag,
+            'VEGAmag': u.mag,
+            'ABmag': u.ABmag
+        }
+        self._block_size = len(self._data_units)
 
     @property
     def date(self):
@@ -171,7 +202,7 @@ class Query(object):
 
     @property
     def zpt_table(self):
-        """The results returned by the ACS Zeropoint Calculator. (`astropy.table.Table`)"""
+        """The results returned by the ACS Zeropoint Calculator. (`astropy.table.QTable`)"""
         return self._zpt_table
 
     def _check_inputs(self):
@@ -183,6 +214,9 @@ class Query(object):
             True if all inputs are valid, False if one is not.
 
         """
+        valid_detector = True
+        valid_filter = True
+        valid_date = True
         # Determine the submitted detector is valid
         if self.detector not in self._valid_detectors:
             msg = ('{} is not a valid detector option.\n'
@@ -191,7 +225,7 @@ class Query(object):
                                '\n'.join(self._valid_detectors),
                                self._msg_div))
             LOG.error(msg)
-            return False
+            valid_detector = False
 
         # Determine if the submitted filter is valid
         if (self.filt is not None and
@@ -202,12 +236,15 @@ class Query(object):
                                '\n'.join(self.valid_filters[self.detector]),
                                self._msg_div))
             LOG.error(msg)
-            return False
+            valid_filter = False
 
         # Determine if the submitted date is valid
         date_check = self._check_date()
         if date_check is not None:
             LOG.error('{}\n{}'.format(date_check, self._msg_div))
+            valid_date = False
+
+        if not valid_detector or not valid_filter or not valid_date:
             return False
 
         return True
@@ -235,9 +272,15 @@ class Query(object):
             result = '{} does not match YYYY-MM-DD format'.format(self.date)
         else:
             if dt_obj < self._acs_installation_date:
-                result = ('The observation cannot occur '
+                result = ('The observation date cannot occur '
                           'before ACS was installed ({})'.
                           format(self._acs_installation_date.strftime(fmt)))
+            elif dt_obj > self._extrapolation_date:
+                result = ('The observation date cannot occur after the '
+                          'maximum allowable date, {}. Extrapolations after '
+                          'this date invalid due to high uncertainties in the ' 
+                          'component and throughput files'.
+                          format(self._extrapolation_date.strftime(fmt)))
         finally:
             return result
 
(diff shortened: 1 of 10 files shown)


acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+FILTER PHOTPLAM        PHOTFLAM         STmag  VEGAmag  ABmag
+       Angstrom erg / (Angstrom cm2 s) mag(ST)   mag   mag(AB)
+------ -------- ---------------------- ------- ------- -------
+ F435W   4329.2              3.148e-19  25.155  25.763  25.665
+"""
+
+
+
+
+from astropy.table import Table
```
pllim: Can we separate the imports as documented in https://www.python.org/dev/peps/pep-0008/#imports ?

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+import datetime as dt
+import requests
+from bs4 import BeautifulSoup
+
+
+__taskname__ = "acszpt"
+__version__ = "1.0"
+__author__ = "Nathan Miles"
+__vdate__ = "22-Jan-2019"
+__all__ = ['acszpt']
```
pllim: I don't see any `acszpt` function in here, do you mean this?
```suggestion
__all__ = ['Query']
```

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+                 F606W, F625W, F658N, F660N,
+                 F775W, F814W, F850LP, F892N
+            * SBC:
+                F115LP, F122M, F125LP,
+                F140LP, F150LP, F165LP
+
+    Attributes
+    ----------
+    date : str
+        **Attribute** for storing the user specified date.
```
pllim: No reason to highlight nor mention `**Attribute**`. These are already in the "Attributes" section, should be obvious enough?

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+    """ Class used to interface with the ACS Zeropoints Calculator API.
+
+    Parameters
+    ----------
+    date : str
+        Input date in the following ISO format, YYYY-MM-DD
+    detector : str
+        One of the three channels on ACS: HRC, SBC, or WFC
+    filt : str, optional
+        One of valid filters for the chosen detector
```
pllim: What does it mean when default (`None`) is given? Perhaps this was explained somewhere else but best repeated here as well.

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+class Query(object):
+    """ Class used to interface with the ACS Zeropoints Calculator API.
+
+    Parameters
+    ----------
+    date : str
+        Input date in the following ISO format, YYYY-MM-DD
+    detector : str
```
pllim: Notation for when only specific choices (and not too many of them) are allowed.
```suggestion
    detector : {'WFC', 'SBC', 'HRC'}
```

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+                    'F660N', 'F775W', 'F814W', 'F850LP','F892N'],
+
+            'SBC': ['F115LP', 'F122M', 'F125LP',
+                   'F140LP', 'F150LP', 'F165LP']
+        }
+        self.zpt_table = None
+
+        # Set the private attributes
+        if filt is None:
+            self._url = 'https://acszeropoints.stsci.edu/results_all/?date={}&' \
```
pllim: Try to avoid `\` continuation if possible. Use parenthesis instead.

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+                format(self.date, self.detector, self.filt)
+
+        self._zpts = OrderedDict()
+        self._response = None
+        self._failed = False
+
+    def _check_inputs(self):
+        """ Check the input to ensure they are valid
+
+        Returns
```
pllim: This looks like an incomplete docstring. It seems to return `bool` but that is not documented.

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+        -------
+
+        """
+        valid_date = True
+        valid_detector = True
+        valid_filter = True
+        # Determine the submitted detector is valid
+        if self.detector not in ['HRC','SBC','WFC']:
+            print('Whoops! {} is not a valid detector option.'.
```
pllim: Using `print` is not recommended. Please consider using Python logging system.

Also, is there a reason why not raise an error (e.g., `ValueError`) straight away? It would be easier to debug if traceback points out where the comparison is failing. In addition, checking behavior is different from `_check_date`, which does raise an error. I think this can be simplified and made more consistent.

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+        # Only continute if all the inputs are valid options
+        if all([valid_date, valid_filter, valid_detector]):
+            return True
+        else:
+            return False
+
+    def _check_date(self, fmt='%Y-%m-%d'):
+        """Convenience method for determing if the input date is valid
+        """
+        min_date = dt.datetime(2002, 3, 7)
```
pllim: No point recalculating a constant at method level? Maybe move it up to module level or class level?

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+        try:
+            dt_obj = dt.datetime.strptime(self.date, fmt)
+        except ValueError:
+            return ValueError('Date format does not match YYYY-MM-DD')
```
pllim: Usually an exception is raised, not returned. If you want to return something, I recommend a status code variable rather than Python exception.

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+
+
+from astropy.table import Table
+import astropy.units as u
+from collections import OrderedDict
+import datetime as dt
+import requests
+from bs4 import BeautifulSoup
```
pllim: To make `bs4` and *optional* import as discussed, please do the following:
```suggestion
try:
    from bs4 import BeautifulSoup
    HAS_BS4 = True
except ImportError:
    HAS_BS4 = False
```
Then you can use `HAS_BS4`  where necessary in your class. Similar comment for `requests`. Or you can dictate that from now on, they are compulsory dependencies for `acstools` -- your call.

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+    def _submit_request(self):
+        """ Submit a request to the ACS Zeropoint Calculator
+
+        If an exception is raised during the request an error message is
+        printed. Otherwise, the response is saved in the corresponding
+        attribute.
+        """
+        try:
+            self._response = requests.get(self._url)
+        except requests.exceptions.RequestException as e:
```
pllim: Are you sure it raises an exception at all? http://docs.python-requests.org/en/master/

I thought its behavior is to store failure inside `self._response.status_code`?

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+        results.
+
+        """
+        soup = BeautifulSoup(
+            self._response.content.decode(self._response.encoding),
+            'html.parser'
+        )
+        # Grab all elements in the table returned by the ZPT calc.
+        td = soup.find_all('td')
+        for j in range(6):
```
pllim: Magic number! Why 6?

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+
+            elif key.lower() == 'abmag':
+                data_types.append(float)
+                data_units.append(u.ABmag)
+
+        tab = Table(list(self._zpts.values()),
+                    names=list(self._zpts.keys()),
+                    dtype=data_types)
+
+        # Loop through each column and set the appropriate units
```
pllim: Is this necessary? Does `astropy.table.QTable` handle this for you?

acstools/acszpt.py
```diff
@@ -0,0 +1,329 @@
...
+            else:
+                self._parse_response()
+                self._format_results()
+                return self.zpt_table
+        else:
+            print('Please fix the incorrect input(s)')
+            return None
+
+
+if __name__ == '__main__':
```
pllim: This looks like leftover code for debugging. Please remove. Or you can turn this into a proper test but have Jenkins skip it by default, or use mock to avoid network calls. The latter is preferred but more work.

acstools/acszpt.py
nmiles2718: That's for the RTD pages. For some reason, I could not get autodoc to generate a header over those definitions and so it just looks like a repetition of the keyword args passed to the instance. It correctly parses all public variables and methods, but for some reason will not generate a header like it does for the parameters even though I was following the [numpydoc conventions](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring) for classes. I ended up removing the Methods section in the class docstring because it was redundant with the methods docstring, in case you're wondering why it's not there.

acstools/acszpt.py
nmiles2718: Just for my own edification, is there a distinct advantage to using parenthesis vs `\` or is this just a stylistic request?

acstools/acszpt.py
nmiles2718: I've never used the logging system, any suggestions on where to find good examples of the implementation? 

The point of these checks is not for the user to debug issues with the code. It's for me to know whether or not they've instantiated the object with valid inputs. If any of the inputs are invalid, the user is informed of which inputs were invalid and what the acceptable inputs are. For example:

An incorrect detector is supplied but the date is correct:
```console 
In [6]: q = acszpt.Query(detector='UVIS', date='2004-02-01')                              

In [7]: q.fetch()                                                                         
Whoops! UVIS is not a valid detector option.
Please choose one of the following:
HRC
SBC
WFC
-------------------------------------------------------------------------------
Please fix the incorrect input(s)
```
An incorrect detector and date are supplied:
```console 
In [8]: q = acszpt.Query(detector='UVIS', date='1999-02-01')                              

In [9]: q.fetch()                                                                         
Whoops! UVIS is not a valid detector option.
Please choose one of the following:
HRC
SBC
WFC
-------------------------------------------------------------------------------
Whoops! The observation cannot occur before ACS was installed (2002-03-07)
-------------------------------------------------------------------------------
Please fix the incorrect input(s)
```
An incorrect date and filter are supplied with a valid detector:
```console
In [10]: q = acszpt.Query(detector='WFC', date='1999-02-01',filt='F140LP')                 

In [11]: q.fetch()                                                                         
Whoops! F140LP is not a valid filter for WFC
Please choose one of the following:
F435W
F475W
F502N
F550M
F555W
F606W
F625W
F658N
F660N
F775W
F814W
F850LP
F892N
-------------------------------------------------------------------------------
Whoops! The observation cannot occur before ACS was installed (2002-03-07)
-------------------------------------------------------------------------------
Please fix the incorrect input(s)
```
Finally, if the detector is invalid and a filter is supplied, the code will skip the filter checking because the detector is not valid and I am not going code every filter for every detector on HST into this. 
```console
In [12]: q = acszpt.Query(detector='UVIS', date='1999-02-01',filt='F140LP')                

In [13]: q.fetch()                                                                         
Whoops! UVIS is not a valid detector option.
Please choose one of the following:
HRC
SBC
WFC
-------------------------------------------------------------------------------
Whoops! The observation cannot occur before ACS was installed (2002-03-07)
-------------------------------------------------------------------------------
Please fix the incorrect input(s)
```

In light of that, I don't see the point of raising errors. I am all for simplification of this code, but enforcing consistency for the sake of consistency seems pointless. I can modify the `try/except` in `_date_check` to simply return a `str` instead of the ValueError if it's _truly_ necessary though.

acstools/acszpt.py
nmiles2718: yeah, this is a good suggestion. I've moved it to the class level and renamed it to be something more useful (`_acs_installation_date`)

acstools/acszpt.py
nmiles2718: I _think_ the only case that it will raise an exception for is when the server is down and it exceeds that max retries. 
```console
In [4]: q = acszpt.Query(detector='WFC', date='2004-02-01')                               

In [5]: q.fetch()                                                                         
Out[5]: 
<Table length=13>
FILTER PHOTPLAM        PHOTFLAM         STmag  VEGAmag  ABmag 
       Angstrom erg / (Angstrom cm2 s) mag(ST)   mag   mag(AB)
 str6  float64         float64         float64 float64 float64
------ -------- ---------------------- ------- ------- -------
 F435W   4328.8               3.07e-19  25.182   25.79  25.692
 F475W   4745.2              1.786e-19   25.77  26.173  26.081
 F502N   5023.0              5.148e-18  22.121  22.388  22.308
 F550M   5581.3              3.921e-19  24.917  24.844  24.875
 F555W   5360.1              1.927e-19  25.688  25.734  25.734
 F606W   5919.3              7.694e-20  26.685  26.423  26.515
 F625W   6311.1              1.172e-19  26.227  25.749  25.919
 F658N   6584.0              1.947e-18  23.176  22.394  22.776
 F660N   6599.4              5.096e-18  22.132  21.441  21.726
 F775W   7693.5              9.849e-20  26.417  25.284  25.678
 F814W   8047.1              6.957e-20  26.794   25.53  25.958
F850LP   9036.5              1.491e-19  25.966  24.352  24.878
 F892N   8914.8              1.477e-18  23.477  21.923  22.418

In [6]: q._url                                                                            
Out[6]: 'http://127.0.0.1:5000/results_all/?date=2004-02-01&detector=WFC'
```
Now I shut down the development server and attempt the same query, 
```console
In [7]: q.fetch()                                                                         
HTTPConnectionPool(host='127.0.0.1', port=5000): Max retries exceeded with url: /results_all/?date=2004-02-01&detector=WFC (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x1c216757b8>: Failed to establish a new connection: [Errno 61] Connection refused',))
-------------------------------------------------------------------------------
The query failed! Check your input args and if the error persists submit a ticket to the ACS Help Desk at hsthelp.stsci.edu with the error message displayed above.
```
I'm glad you pointed this out because for some reason I thought the `requests` module was a native python module. I switched everything to use `urllib.request` and its corresponding errors module `urllib.error` (no need for unnecessary dependencies). The result is the same in that if the server is down it will raise an exception stating max retries, so the code is unchanged.

acstools/acszpt.py
nmiles2718: There are 6 columns in the parsed html table, so I have to loop through in blocks of 6. Couldn't really come up with a cleaner way to do this that didn't involve hard coding that value.

acstools/acszpt.py
nmiles2718: Which part are you referring to? The long block of `if/elif` statements is required to specify the units and datatypes for each column. The only other option is to hard code a list with those values, but we run the risk of the hard coded lists not being in the same order as the columns, i.e. we could potentially say the `FILTER` column is comprised of `float` with `u.ABmag` or something.

acstools/acszpt.py
nmiles2718: If you're referring to the `if __name__ == '__main__':`, then yeah it is leftover for debugging and I'll remove it

setup.py
```diff
@@ -66,7 +66,8 @@
     entry_points=entry_points,
     install_requires=[
         'astropy>=1.1',
-        'numpy'
+        'numpy',
+        'bs4'
```
pllim: What about `requests`?

acstools/acszpt.py
pllim: There is some basic logging in
https://github.com/spacetelescope/acstools/blob/88529aadf4ab0348ff0ad25913da14e65517f522/acstools/acs_destripe.py#L92-L94

setup.py
nmiles2718: I replaced `requests` with `urllib`

pllim: Oh, ok. Yes, I remember reading about that earlier. Sorry for the noise!

pllim: Hmm... Are you sure it is not `'beautifulsoup4'` as per PyPI listing? https://pypi.org/project/beautifulsoup4/

pllim:
```
$ conda list soup
# Name                    Version                   Build  Channel
beautifulsoup4            4.6.3                     <pip>
```

nmiles2718: Either works. The author has a dummy package called `bs4` that is aliased to `beautifulsoup4`. For some reason they decided to call it `beautifulsoup4`, yet if you try to import a package by that name, it will raises an import error. It can only be imported with `import bs4`, regardless of the route of installation (i.e. `pip install bs4` or `pip install beautifulsoup4`). /shrug

pllim: @jhunkeler said best to make it something that `conda` also recognize.

pllim:
@nmiles2718 , let's discuss further over at nmiles2718/acstools#1. Thanks!

acstools/acszpt.py
```diff
@@ -0,0 +1,379 @@
...
+            msg = ('{} is not a valid detector option.\n'
+                   'Please choose one of the following:\n{}\n'
+                   '{}'.format(self.detector,
+                               '\n'.join(self._valid_detectors),
+                               self._msg_div))
+            LOG.error(msg)
+            valid_detector = False
+
+        # Determine if the submitted filter is valid
+        if (self.filt is not None and
```
pllim: You need to add back the `valid_detector` check here, otherwise you will get `KeyError`. I took that out in my refactoring, which you reverted to.

acstools/acszpt.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        result = None
+        try:
+            dt_obj = dt.datetime.strptime(self.date, fmt)
+        except ValueError:
+            result = '{} does not match YYYY-MM-DD format'.format(self.date)
+        else:
+            if dt_obj < self._acs_installation_date:
+                result = ('The observation date cannot occur '
+                          'before ACS was installed ({})'.
```
pllim: I am not a fan of separating the `.` and the `format()` like this as it is easy to miss in debugging.

acstools/acszpt.py
```diff
@@ -0,0 +1,379 @@
...
+                          (int(len(td) / self._block_size), self._block_size))
+        # Create the QTable, note that sometimes self._response will be empty
+        # even though the return was successful; hence the try/except to catch
+        # any potential index errors. Provide the user with a message and
+        # set the zpt_table to None.
+        try:
+            tab = QTable(data[1:, :], names=data[0],
+                       dtype=[str, float, float, float, float, float])
+        except IndexError as e:
+            msg = ('{}\n{}\nHmm, there was an issue parsing the request. '
```
pllim: There is no need to insert sounds like "hmm" and "whoops".
```suggestion
            msg = ('{}\n{}\nThere was an issue parsing the request. '
```

acstools/acszpt.py
nmiles2718: Good catch!

pllim:
I think we have beaten this horse to death. 😅  Thanks for your patience and cooperation!

PR #81, opened 2019-02-27
MNT: Remove PY2 support

### LINKED ISSUE #65: Retire Python 2
Python 2 will not be maintained past Jan 1, 2020 (see https://pythonclock.org/). Please remove all Python 2 compatibility and move this package to Python 3 only. 

For conda recipe (including `astroconda-contrib`), please include the following to prevent packaging it for Python 2 (https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html?preprocessing-selectors#skipping-builds): 

```
build:
   skip: true  # [py2k]
``` 

Please close this issue if it is irrelevant to your repository. This is an automated issue. *If this is opened in error, please let pllim know!*

pllim:
Fix #65 . Also removed mention of PyRAF since it is only distributed in Python 2. This should be the last PR needed before the next release.

### DIFF
--- a/acstools/__init__.py
+++ b/acstools/__init__.py
@@ -6,8 +6,6 @@ Utility and library functions used by these tasks are also included in this
 module.
 
 """
-from __future__ import absolute_import, print_function
-
 try:
     from .version import *
 except ImportError:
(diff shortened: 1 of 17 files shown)


pllim:
I have not used the TEAL interface in years. It is buried in `stsci.tools` and not maintained. Oh dear... We have two options:

* Fix it up. (Takes longer)
* Remove TEAL support in `acstools`. (Faster to do but bad news if your users love TEAL.)

PR #107, opened 2020-02-25
[EXP] Use astropy config

pllim:
**DO NOT MERGE**

I don't know why this is needed. Just brainstorming as part of spacetelescope/stsci.tools#94.


pllim:
Not needed anymore. See #108

PR #108, opened 2020-03-18
MNT: Remove TEAL support

### LINKED ISSUE #106: MNT: Remove TEAL
If TEAL interface is not used here, we can safely remove it. @tddesjardins will discuss with @nagrogin

pllim:
Fix #106 

**TODO**

- [x] Get approval from @tddesjardins


acstools/acs_destripe_plus.py
```diff
@@ -18,20 +18,14 @@
 Examples
 --------
 
-In Python without TEAL (recommended):
+In Python (recommended):
```
nmiles2718: Do we need the "(recommended)" part? With the removal of TEAL, there is no other options from within acstools and so this is redundant.

acstools/acs_destripe.py
```diff
@@ -22,19 +22,13 @@
 Examples
 --------
 
-In Python without TEAL (recommended):
+In Python (recommended):
```
nmiles2718: Again, no need to include "(recommended)" as using it via Python is the only option

acstools/acs_destripe_plus.py
pllim: I think we do because there is also the CLI option?

acstools/acs_destripe.py
pllim: Same as above. It has a CLI.

acstools/acs_destripe_plus.py
nmiles2718: We only ever mention running `calacs.e` on the RTD pages and the ACS DataHandbook always references the `acstools` interface. If you're tied to it, I'm fine with leaving it but it just seems weird to say something is recommended but never mention the other option that its recommended over.

pllim: Haha... okie dokie. Good point. I'll remove it.

pllim:
@nmiles2718 , if it looks good to you now, please approve. Thanks!

pllim:
Jenkins failure is due to Artifactory being down and not related to this PR.

Thanks for the review!

PR #109, opened 2020-03-23
acsphotcte

### LINKED ISSUE #104: Enhancement: ACS Photometric CTE API
I am currently in the process of finalizing a serverless approach to the [ACS Photometric CTE Webtool](https://acsphotometriccte.stsci.edu). The idea is to provide users with a programmatic interface in a manner that is similar to what was done for the [ACS Zeropoints Calculator](https://acszeropoints.stsci.edu) and the `acszpt` module (#78).  

The details of the implementation and services used may be found [here](https://innerspace.stsci.edu/display/AWS/ACS+Webtool+on+AWS+public+Cloud). 

Assuming you have some catalog (`tb`) containing the fluxes, number of y transfers, and the local sky backgrounds for N sources in an FLT image then to compute the photometric CTE corrections, you would do something along the lines of:
```python
from acstools import acsphotcte

photcteapi = acsphotcte.PhotCTEAPI()
corrections = photcteapi.compute_corrections(
    radius=3,
    mjd=58000,
    fluxes=tb['fluxes'],
    ytransfers=tb['ytransfers'],
    local_skys=tb['local_skys']
)
```

nmiles2718:
This PR contains all of the updates required to support the frontend of the ACS Photometric CTE API referenced in issue #104.

The backend for this API has been approved by the Cloud Application Review Board and a complete review of everything may be found on the innerspace page, [ACS Webtool on AWS public cloud](https://innerspace.stsci.edu/display/AWS/ACS+Webtool+on+AWS+public+Cloud+by+Nathan+Miles+-+INS).

The plan here is to deploy the application backend into production, update the API URL and public key information in `acsphotcte.PhotCTEAPI()`, then merge this PR and release the tool to the public. The targeted release date is March 26th. 

Closes #104


pllim:
Please fix PEP 8 and do a rebase/squash to get rid of the merge commits. Thanks!

pllim:
Note to self: This needs a new release to be done after merge.

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+...    fluxes=fluxes
+...)
+>>> print(cte_corrections)
+"""
+from collections import Iterable
+import json
+import logging
+
+import numpy as np
+import requests
```
pllim: `requests` should be an optional dependency. It is not used anywhere else in this package (that I can find).

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+
+__taskname__ = "acsphotcte"
+__author__ = "Nathan Miles"
+__version__ = "1.0"
+__vdate__ = "16-Mar-2020"
+
+__all__ = ['PhotCTEAPI']
+
+
+class PhotCTEAPI():
```
pllim:
```suggestion
class PhotCTEAPI:
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+        self._cte_corrections = None
+        self._corrected_fluxes = None
+
+    @property
+    def corrected_fluxes(self):
+        """CTE corrected fluxes"""
+        return self._corrected_fluxes
+
+    @corrected_fluxes.setter
+    def corrected_fluxes(self, value):
```
pllim: Is there no need for value validation here?

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+        self._corrected_fluxes = value
+
+    @property
+    def cte_corrections(self):
+        """Computed CTE corrections"""
+        return self._cte_corrections
+
+    @cte_corrections.setter
+    def cte_corrections(self, value):
```
pllim: Is there no need for value validation here?

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+__taskname__ = "acsphotcte"
+__author__ = "Nathan Miles"
+__version__ = "1.0"
+__vdate__ = "16-Mar-2020"
+
+__all__ = ['PhotCTEAPI']
+
+
+class PhotCTEAPI():
+    """Convenience class for handling queries to ACS Photometric CTE API
```
pllim:
```suggestion
    """Convenience class for handling queries to ACS Photometric CTE API.
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+            The original dictionary with all of the inputs converted to their
+            python counterparts for JSON serialization
+        """
+        for key in inputs.keys():
+            # Check if the input is Iterable and not a list  (e.g. np.array)
+            # We also need to make sure the input is not a str, since str are
+            # Iterable in Python.
+            if isinstance(inputs[key], Iterable) and \
+                    not isinstance(inputs[key], list) and \
+                    not isinstance(inputs[key], str):
```
pllim:
```suggestion
            if (isinstance(inputs[key], Iterable) and
                    not isinstance(inputs[key], (list, str))):
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+                # Convert the array to a python list
+                inputs[key] = list(inputs[key])
+
+                # Check if the first element of the input array is a float
+                if not isinstance(inputs[key][0], float):
+                    LOG.info('Converting elements..')
+                # Convert the array of elements to floats
+                    inputs[key] = list(map(float, inputs[key]))
```
pllim: Doesn't this effectively modify `inputs` in-place? Do you still have to return it below?

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+
+        local_sky : float, list
+            Local sky background for each source
+
+        fluxes : float, list
+            Computed source flux, in electrons, for 3 or 5 pixel apertures.
+
+        Returns
+        -------
+
```
pllim: What does it return?

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+            'fluxes': fluxes
+        }
+        inputs = self._check_inputs(**inputs)
+        if not isinstance(inputs, dict):
+            LOG.info(f'Did you forget to supply the {inputs}?')
+            return
+        content = self._query(inputs, api_credentials)
+
+        # Update the attribute and return the value
+        self.cte_corrections = np.array(content['deltamag'])
```
pllim: No handling of failed query?

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+            self,
+            radius=None,
+            ytransfers=None,
+            mjd=None,
+            local_skys=None,
+            fluxes=None,
+            api_id=None,
+            api_key=None
+    ):
+        """Get the CTE corrected FLT photometry
```
pllim:
```suggestion
        """Get the CTE corrected FLT photometry.
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+
+        Parameters
+        ----------
+        radius : {3, 5}
+            Aperture radius used to extract photometry
```
pllim:
```suggestion
            Aperture radius used to extract photometry.
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+
+        ytransfers : float, list
+            Number of y transfers. For chip 2 on WFC this is just the Y value,
```
pllim:
```suggestion
            Number of row transfers. For chip 2 on WFC, this is the Y value.
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+            for chip 1 it is 2048 - Y value.
```
pllim:
```suggestion
            For chip 1, it is ``2048 - Y`` value.
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+
+        mjd : float
+            MJD date of the observation
```
pllim: Not everyone think in MJD.  I wonder if it is better to support `YYYY-MM-DD` but that discussion is out of scope here.
```suggestion
            MJD date of the observation.
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+
+        local_sky : float, list
+            Local sky background for each source
```
pllim: In what unit?
```suggestion
            Local sky background for each source.
```

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+        if not isinstance(inputs, dict):
+            LOG.info(f'Did you forget to supply the {inputs}?')
+            return
+        content = self._query(inputs, api_credentials)
+
+        # Update the attribute and return the value
+        self.cte_corrections = np.array(content['deltamag'])
+        return self.cte_corrections
+
+    def correct_photometry(
```
pllim: Not all the inputs are explained in docstring. Why?

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+pixel aperture radii are supported. The model is designed to compute the
+CTE losses as a function of the number of Y transfers, sky background,
+source flux, and observation date.
+
+
+Usage Guidelines
+----------------
+As this service is hosted on AWS, there are some guidelines that should be
+followed when utilizing this tool. For each call to
+:py:meth:`PhotCTEAPI.correct_photomerty`, AWS will assemble the compute resources,
```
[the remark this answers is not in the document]
nmiles2718: Hmm, interesting. It worked in the local build, but I'll investigate further

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+sources in a single request.
+
+
+Example
+-------
+Get the CTE corrected FLT photometry for a list of 1000 sources:
+
+>>> import numpy as np
+>>> from acstools import acsphotcte
+>>> ytransfers = 2048 * np.random.random(size=1000)
```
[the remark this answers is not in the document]
nmiles2718: The only number I haven't pulled out of a hat is 2048 and that is because each ACS/WFC chip has a maximum of 2048 y transfers. Everything else is completely arbitrary and I don't think they need any explanation.

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+>>> fluxes = 20000*np.random.random(size=len(ytransfers))
+>>> local_skys = 80*np.random.random(size=len(ytransfers))
```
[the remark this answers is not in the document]
nmiles2718: It will always be 1,000. I just did that so if I could see how well it handled 1 vs 10 vs 100 vs 1,000 vs 10,000 vs 100,000 without having to change it in 3 different places every time.

acstools/acsphotcte.py
```diff
@@ -0,0 +1,289 @@
...
+>>> mjd = 59341.
+>>> radius = 3.
+>>> cte_corrections = acsphotcte.correct_photometry(
+...    radius=radius,
+...    ytransfers=ytransfers,
+...    mjd=mjd,
+...    local_skys=local_skys,
+...    fluxes=fluxes
+...)
+>>> print(cte_corrections)
```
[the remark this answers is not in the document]
nmiles2718: I'll add it in

acstools/acsphotcte.py
nmiles2718: The modified `inputs` is only present in the scope of that method right?

acstools/acsphotcte.py
nmiles2718: Great catch!

acstools/acsphotcte.py
pllim: `inputs` is a dictionary, so I thought it would be mutable. You can check for yourself.

acstools/acsphotcte.py
nmiles2718: The `api_key` and `api_id` arguments are only to be used for development purposes. Is there a proper way to handle this other than just documenting that they're to be used for that?

acstools/acsphotcte.py
nmiles2718: I don't think so. If these attributes are set, that means everything went as expected. Otherwise the remain `None`.

acstools/acsphotcte.py
nmiles2718: Same as above

acstools/acsphotcte.py
nmiles2718: I think I still need to return it.
```console
In [6]: def change_values(**inputs):
   ...:     for key in inputs.keys():
   ...:         print(f'Changing the value {key}')
   ...:         inputs[key]='changed!'
   ...:
   ...: def test():
   ...:     d = {'a':[1,2,3,4],
   ...:         'b':'original'}
   ...:
   ...:     change_values(**d)
   ...:     print(d)
   ...:

In [7]: test()
Changing the value a
Changing the value b
{'a': [1, 2, 3, 4], 'b': 'original'}

In [8]: d = {'a':1234,'b':'1234'}

In [9]: change_values(**d)
Changing the value a
Changing the value b

In [10]: d
Out[10]: {'a': 1234, 'b': '1234'}```

File: acstools/acsphotcte.py (hunk 3 above)
Comment (nmiles2718) [reply acstools/acsphotcte.py]: Also, I realized that `corrected_fluxes` should really be `corrected_magnitudes` just an FYI before the next commit.

File: acstools/acsphotcte.py (hunk 8 above)
Comment (nmiles2718) [reply acstools/acsphotcte.py]: I'm going to just remove this method as it's not really necessary.

File: acstools/acsphotcte.py (hunk 1 above)
Comment (nmiles2718) [reply acstools/acsphotcte.py]: Good catch! Haven't had to add optional dependencies before, so I want to double it's through the `extras_require` keyword in `setup()` right? So something like,
```python
install_requires=[
        'astropy>=2',
        'numpy',
        'beautifulsoup4'
    ],
extras_require={'request':'requests'},
ests_require=['pytest'],
```

acstools/acsphotcte.py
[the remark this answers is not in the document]
pllim: I disagree. Arbitrary numbers are confusing without context. Your goal in this example is to let people who have no experience with ACS to be able to use your product.

acstools/acsphotcte.py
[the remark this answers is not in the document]
pllim: Why not do a `n_sample = 1000` and then re-use `n_sample`?

acstools/acsphotcte.py
pllim: Not only that; It should only be imported inside your method that uses it, not at the module level.

acstools/acsphotcte.py
pllim: Ah, maybe because you passed it in with `**`. Nvm then. Thanks for checking.

acstools/acsphotcte.py
nmiles2718: Ok, my only question is why not add it to requirements? If we want the module to be useful, then people need to have requests installed or nothing will happen. We had a similar scenario with BeautifulSoup and the `acszpt.py` module. No other module in `acstools` uses BeautifulSoup by `acszpt.py` is useless without that package installed.

acstools/acsphotcte.py
pllim: At the very least, state that they are for development purpose only? Although, API doc is usually geared for developers to read. For instance, what do you want the next maintainer to know about these options?

acstools/acsphotcte.py
pllim: Good point about the `bs4` requirement. I guess we already set a precedent. If the list keeps growing though, I need to rethink this as not to bloat up the compulsory dependencies of `acstools`.

nmiles2718: In the future if it's necessary, we could replace the `requests` module with `urllib.request`. For now, I've added `requests` to the dependencies.

acstools/acsphotcte.py
```diff
@@ -0,0 +1,282 @@
...
+                    # TypeError during the conversion, print the info and
+                    # return the key it failed on
+                    LOG.error(e)
+                    return key
+                else:
+                    if key == 'radius' and inputs[key] not in [3., 5.]:
+                        LOG.error('Submitted radii not supported...')
+                        return key
+        # the last step is to ensure all the iterable inputs are the same length
+        nl = "\n"
```
pllim: If you ever plan to support Windows...
```suggestion
        nl = os.linesep
```

PR #117, opened 2020-03-31
[quickfix] acsphotcte

nmiles2718:
- Adds error handling to ensure the supplied MJD date occurs after SM4.
- Fixes typo in documentation where ACS ISR 2012-02 is referenced instead of ACS ISR 2015-05.

### DIFF
--- a/acstools/acsphotcte.py
+++ b/acstools/acsphotcte.py
@@ -4,9 +4,9 @@ The ACS Photometric CTE API is a programmatic interface for the
 `ACS Photometric CTE Webtool <https://acsphotometriccte.stsci.edu>`_.
 The API is a cloud-based service that employs a serverless approach on AWS
 with API Gateway and Lambda to compute the photometric CTE corrections
-using the model described in ACS ISR 2012-02.
+using the model described in ACS ISR 2012-05.
 The model corrects ACS/WFC aperture photometry extracted from FLT images for
-CTE losses using the most recent formula from Chiaberge's ISR 2012-05. It is
+CTE losses using the most recent formula described in Chiaberge's ISR 2012-05. It is
 only calibrated for photometry obtained after SM4 in May 2009. For pre-SM4
 data, please see Chiaberge et al. 2009 (ISR 2009-01), or use pixel-based
 CTE-corrected files obtained from MAST. Currently,  only 3 and 5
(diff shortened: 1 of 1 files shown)


acstools/acsphotcte.py
```diff
@@ -167,6 +167,9 @@ def _check_inputs(self, **inputs):
             it returns the corresponding key. If the iterable inputs are not
             all the same length, it returns `None`.
         """
+        # MJD date of Servicing Mission 4, 2009-05-11
+        # This is the same date used for the CTE Webtool
+        sm4 = 54962.0
```
pllim: Why is this different from the one that Norman defined in `acs_destripe`?

https://github.com/spacetelescope/acstools/blob/4284f9575943ab1b520e2aa83d27ace4d97d1e9d/acstools/acs_destripe.py#L75

acstools/acsphotcte.py
```diff
@@ -210,6 +213,10 @@ def _check_inputs(self, **inputs):
             )
             LOG.error(msg)
             return
+        # Check to see if the date occurs before SM4
+        if inputs['mjd'] < sm4:
+            LOG.error('Observation MJD must occur after SM4, 54962')
```
pllim: Maybe use f-string to insert the SM4 value instead of hardcoding.

acstools/acsphotcte.py
```diff
@@ -37,11 +37,11 @@
 >>> import numpy as np
 >>> from acstools import acsphotcte
```
pllim: Might as well define a variable?
```suggestion
>>> from acstools import acsphotcte
>>> n_sample = 1000
```

acstools/acsphotcte.py
```diff
@@ -37,11 +37,11 @@
...
 >>> ytransfers = 2048 * np.random.random(size=1000)
```
pllim:
```suggestion
>>> ytransfers = 2048 * np.random.random(size=n_sample)
```

acstools/acsphotcte.py
```diff
@@ -37,11 +37,11 @@
...
->>> fluxes = 20000*np.random.random(size=len(ytransfers))
+>>> fluxes = 20000*np.random.random(size=1000)
```
pllim:
```suggestion
>>> fluxes = 20000*np.random.random(size=n_sample)
```

acstools/acsphotcte.py
```diff
@@ -37,11 +37,11 @@
...
 >>> magnitudes = -2.5 * np.log10(fluxes)
 >>> print(magnitudes[:5])
 [-10.75088228  -9.56321561 -10.10571966  -9.01893015 -10.48463087]
->>> local_skys = 80*np.random.random(size=len(ytransfers))
+>>> local_skys = 80*np.random.random(size=1000)
```
pllim:
```suggestion
>>> local_skys = 80*np.random.random(size=n_sample)
```

acstools/acsphotcte.py
nmiles2718: Not sure, the value I use corresponds to the launch date for SM4. The value listed in `acs_destripe.py` is 5 days later on May 16 2009. The discrepancy shouldn't be any issue and so I am happy to conform to the date listed in that file.

pllim: Would be nice if ACS Team can agree on a single value. Please see #118.

PR #119, opened 2020-03-31
Standardizes the SM4 MJD date and updates affected modules

### LINKED ISSUE #118: MNT: Standardize SM4 MJD
There are a few modules that define their own SM4 MJD values and they don't even agree with each other. I propose we define the one true SM4 MJD (maybe under `utils` or something) and use it and only it throughout the package.

nmiles2718:
- After conversation with the ACS Team, the agreed upon MJD date for SM4 is 54967. This date corresponds to the EVA for ACS repair during SM4. 

This PR closes #118

### DIFF
--- a/acstools/utils_calib.py
+++ b/acstools/utils_calib.py
@@ -22,7 +22,7 @@ __all__ = ['extract_dark', 'extract_flash', 'extract_flatfield',
            'check_overscan', 'SM4_MJD']
 
 # The MJD date of the EVA during SM4 to restore ACS/WFC and ACS/HRC.
-# This value is also defined in header file, acs.h, for us by calacs.e in hstcal
+# This value is also defined in header file, acs.h, for use by calacs.e in hstcal
 SM4_MJD = 54967.0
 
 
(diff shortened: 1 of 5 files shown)


acstools/utils_calib.py
```diff
@@ -21,6 +21,9 @@
            'from_lt', 'hdr_vals_for_overscan', 'check_oscntab',
            'check_overscan']
```
pllim:
```suggestion
           'check_overscan', 'SM4_MJD']
```

acstools/utils_calib.py
```diff
@@ -21,6 +21,9 @@
...
 
+# The MJD date of the EVA during SM4 to restore ACS/WFC and ACS/HRC
```
pllim: Maybe also add a note that this value is also defined in https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/include/acs.h#L88 , as pointed out by @mdlpstsci .

acstools/utils_calib.py
```diff
@@ -19,7 +19,11 @@
 __all__ = ['extract_dark', 'extract_flash', 'extract_flatfield',
            'from_irafpath', 'extract_ref', 'find_line', 'get_corner', 'get_lt',
            'from_lt', 'hdr_vals_for_overscan', 'check_oscntab',
-           'check_overscan']
+           'check_overscan', 'SM4_MJD']
+
+# The MJD date of the EVA during SM4 to restore ACS/WFC and ACS/HRC.
+# This value is also defined in header file, acs.h, for us by calacs.e in hstcal
```
pllim, line 25:
```suggestion
# This value is also defined in header file, acs.h, for use by calacs.e in hstcal
```

PR #46, opened 2017-11-09
Enable CRREJ on acs_destripe_plus products

pllim:
Fix #35 -- Enable CRREJ on `acs_destripe_plus` products. Needed to add a new function named `crrej_plus()` due to the recursive nature of `acs_destripe_plus()`, which was meant to process individual exposures, not combine them.

Fix #66 

Bonus: Removed unexpected `_RULES_` section warning when using TEAL interface. Some minor PEP 8 clean-ups on the module.

c/c @jryon , @nmiles2718 , and @tddesjardins

PR #120, opened 2020-03-31
ENH: Enable CRREJ on acs_destripe_plus products, keep tmp files

### LINKED ISSUE #35: acs_destripe_plus cannot produce CRJ/CRC
As reported by @dborncamp , there is currently no way for `acs_destripe_plus` to also generate CRJ/CRC files. This is because while it tries to mimic `calacs.e` (but with the standalone destriping in between `acsccd` and `acscte`), it does not do so completely. It does not invoke `acsrej` between `acscte` and `acs2d` (see http://www.stsci.edu/hst/acs/documents/handbooks/currentDHB/acs_Ch34.html).

Attached is code from Dave Borncamp to work around this limitation:  https://gist.github.com/pllim/ef0b8bc20b9b9e106913e82d8acb8297

### LINKED ISSUE #66: no option in acs_destripe_plus to retain BLV_TMP/BLC_TMP files
`acs_destripe_plus` does not have an option to retain BLV_TMP/BLC_TMP files. Such a feature would be useful for users running the pixel-based CTE forward model on subarray images, since that step should be run after bias correction but before dark correction and flat-fielding.

pllim:
This is basically #46 without the TEAL stuff.

Fix #35 
Fix #66

We thought `acs2d` was removing input files on the HSTCAL side but I don't think that is the case. I actually don't see the disappearing `blv_tmp` reported by @nmiles2718 in https://github.com/spacetelescope/acstools/pull/46#issuecomment-416599585 when I run this myself on a RAW file I have at hand.

**TODO**

- [ ] Get approval from ACS Team after they test this PR out. cc @jryon


pllim:
RTD build got stuck. I'll leave it be until I rebase, which I am sure I'll need to do at some point.

nmiles2718:
@pllim I'm not convinced that the files are being saved. I tested the changes on this branch and the intermediate files were still deleted. I captured the output via screenshots to demonstrate the behavior and combined them into a single PDF. I will send you a copy via slack.

nmiles2718:
### Steps to reproduce the behavior

Note: All data are located on my personal central storage and I only list the relative paths here.

datadir: /acs_destripe_test/fitsdata/
- There are clean copies of each file and two subdirectories. One subdirectory contains multiple RAW files and the other contains a single RAW file.

Processing a single RAW:
```console
cd /acs_destripe_test/fitsdata/single_raw/

ipython

from acstools import acs_destripe_plus

# check the version and make sure it is 0.4.3
acs_destripe_plus.__version__

# process the file
acs_destripe_plus.destripe_plus('*raw.fits',cte_correct=False, keep_intermediate_files=True)
```

Processing multiple RAWs:
```console
cd /acs_destripe_test/fitsdata/multiple_raw/

ipython
from acstools import acs_destripe_plus

# check the version and make sure it is 0.4.3
acs_destripe_plus.__version__

# process the file
acs_destripe_plus.destripe_plus('*raw.fits',cte_correct=False, keep_intermediate_files=True)
```

Results:
When processing a single image, the intermediate files are saved. When processing multiple images, the intermediate files are deleted.

jryon:
I performed the same tests as @nmiles2718 on different data, and confirmed his results. No BLV_TMP files remain when running acs_destripe_plus on more than one RAW at a time.

pllim:
Yup, I just fixed the bug in the latest commit. I missed the fact that `destripe_plus` calls itself when `n_input>1` and didn't pass in new option to the recursive call. My bad!

pllim:
@nmiles2718 and @jryon , this is now ready for final review and approval. I can confirm that the disappearing TMP files are not disappearing anymore with the dataset that @nmiles2718 provided. I apologize for any inconvenience caused!

jryon:
@pllim We confirmed that the BLV_TMP files are saved correctly with destripe_plus. We also tested crrej_plus, and came across a couple of issues:

1. Something is wrong with the output filename, and it's not clear if it is the treatment of the `outroot` parameter in this function, or if acsrej or acs2d are not naming the output files correctly. If I set `outroot = 'test'`, then the final combined and processed file is 'test_crj_tmp_flt.fits'. If I set `outroot = 'test_crj'`, the final file is 'test_crj_crj_tmp_flt.fits'. 
2. It would be great if the input files could also be of the form '*_blv_tmp.fits', rather than requiring that the input be a list.

pllim:
@jryon , thanks for checking! Is it possible to get the dataset you are testing `crrej_plus` with?

pllim:
I was under the impression that ACS2D is smart enough to rename `crj_tmp` to `crj` but turns out that is not true:

```
CALBEG*** ACS2D -- Version 10.2.1 (25-Sept-2019) ***
Begin    01-Apr-2020 17:40:22 EDT
Input    zzz_crj_tmp.fits
Output   zzz_crj_tmp_flt.fits
```

This might be a bug in HSTCAL.

pllim:
@jryon and @nmiles2718 , I think I have addressed all the concerns at https://github.com/spacetelescope/acstools/pull/120#issuecomment-607480672 . You might also want to test this with `PCTECORR=PERFORM`. Please let me know what you think now. Thanks for your patience!

jryon:
@nmiles2718 can no longer review this PR, so @pllim, can you make me the reviewer? I'll be following up on this today.

pllim:
Done. Thanks for reviewing, @jryon !

jryon:
1. I tested `destripe_plus` with `cte_correct = True` and `keep_intermediate_files = True` on 2K subarray data. All of the intermediate files were retained as expected (BLV_TMP, BLC_TMP) and the final files had the correct names (FLT, FLC). 
2. I tested `crrej_plus` with the BLV_TMPs, and the output file was correctly named (CRJ). There was still a `crj_tmp.tra` file in the directory, however.
3. Running `crrej_plus` on the BLC_TMPs did not produce the expected output file, it was named `_crc_tmp_flt.fits`. There's still a `crc_tmp.tra` file in the directory.

(I didn't test this on other subarrays (new mode 1K and 512) because I discovered the PCTETAB header keyword is not populated correctly in my data. I'll have to follow up on that with other people.)

pllim:
Thanks! No. 3 is definitely a bug, as I guessed wrongly how that HSTCAL bug is manifesting for BLC input; I'll fix it.

As for `crc_tmp.tra`, don't you want to keep it? It contains processing log for the temporary file. Or do you want me to rename/remove it? Please advise.

jryon:
I'm ok with retaining TRA files, but does there need to be a separate TRA file for the temporary file? I noticed that if `outroot='test'`, then I get both `test.tra` and `test_crc_tmp.tra`. When running `destripe_plus`, I only have one TRA file for each RAW file (like `jd5702kcq.tra`).

pllim:
The temporary files are automatically generated by different C executables that we're calling. I can try renaming it before calling the next step to see if it appends, but it might also just crash with "file already exist" error (not sure, I'll have to check).

But if the case is the latter, I feel like manually concatenating them here on the Python side is getting out of scope.

jryon:
Ah I see. Given that they're automatically generated by the C code, I'm fine with leaving them as is. Thanks for the explanation!

pllim:
Re: https://github.com/spacetelescope/acstools/pull/120#issuecomment-610585555 -- @jryon , I believe I have addressed your concerns (both the CRC and TRA names). Please try again with the updated PR branch.

p.s. For completeness, please also report back that the output files are scientifically viable.

jryon:
All of the filenames now appear as expected. The data appear scientifically viable. 

I noticed a couple of header keywords have issues. In the CRC:
1. In extension 0, the keyword FILENAME is `test_crc_tmp_flt.fits`, while it should be `test_crc.fits`. 
2. Also in extension 0, the keyword ROOTNAME is set to `TEST_CRC_TMP`, but should be `TEST`. 3. In extensions 1 through 3, the keyword EXPNAME is `test_crc_tmp` but should be `test`.

In the CRJ:
1. Same as 1 above, but `crj` instead of `crc`. 
2. Also in extension 0, the keyword ROOTNAME is set to `*****`, but should be `TEST`.

I think these issues should also be handled by spacetelescope/hstcal#470, so it's possible the fixes for this can be removed in the future. Sorry for not noticing these earlier!

pllim:
Does it matter what the ROOTNAME of `crx_tmp` files are? They are temporary files.  And yes, they are set within HSTCAL. I bet it has always been this way but you don't see it because HSTCAL is not keeping those files when it runs.

jryon:
My previous comment was about the headers of the final CRJ and CRC files, not the CRX_TMPs.

pllim:
How about now, @jryon ? Thanks for the detailed review and testing!

acstools/acs_destripe_plus.py
```diff
@@ -607,6 +607,15 @@ def crrej_plus(filelist, outroot, keep_intermediate_files=False, verbose=True):
     if os.path.isfile(wrongname) and not os.path.exists(rightname):
         os.rename(wrongname, rightname)
 
+        # Brute-force fixing of final output header.
```
pllim, line 610: Note that, the logic is nested here under the assumption that the header stuff will be fixed at HSTCAL eventually.

pllim:
Thank YOU!

PR #134, opened 2020-09-22
Add ACS polarization tools and data

tddesjardins:
This PR adds tools for ACS polarization calculations as well as some associated data. The primary addition is a `Polarization()` class that takes pol0, pol60, and pol120 measurements as inputs, as well as some associated information (PA_V3, detector, filter), and has methods for computing the Stokes parameters and derived polarization characteristics (fraction, angle).

In addition to the above, I've included hidden methods in the `Polarization()` class for users who might want to use them if, e.g., they don't have ACS count rates, but they have Stokes parameters and want to calculate the fractional polarization/angle. I doubt this will be widely used, but someone may find it useful (and it could be useful for internal calibration work). The primary methods in the class are wrappers around these hidden methods and pass the class attributes as inputs.

In addition, another class called `PolarizerTables()` contains tabulated information from the ACS Data Handbook for the polarizer efficiencies and transmission leak corrections. These are loaded in the `Polarization()` class to get relevant values, but users can also instantiate the `PolarizerTables()` class to get access to AstroPy Table versions of the handbook tables. The data for the tables are contained in the YAML file acstools/data/polarizer_data.yaml.

@pllim please give this a code review. I've asked Dean Hines to do any scientific validation he feels is necessary (don't know his GitHub username, but I'll message him with a link to this PR).


acstools/polarization_tools.py
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env python
+"""
+Toolkit for analyzing HST/ACS polarization data.
+
+CHANGE LOG
```
pllim, line 5: Hmm. Usually I keep the change log as release notes, but we can start utilizing `CHANGES.rst` for real rather than embedding the change log inside a docstring.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,204 @@
...
+----------
+2020-09-22: Creation. (T. Desjardins)
+"""
+
+import numpy as np
+import os
+import yaml
```
pllim, line 12: I can't decide whether to make this package compulsory or optional dependency. Gotta think about it. Regardless, it needs to be defined in `setup.cfg`.

pllim (commented):
From a quick glance (will do more tomorrow):

* Define package data in `setup.cfg`.
* Fix CI failures.
* Add test(s) for this new functionality.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+    >>> print(f'I = {vela_181.stokes_i:.2f}, Q = {vela_181.stokes_q:.2f}, U = {vela_181.stokes_u:.2f}')
+    >>> print(f'Polarization: {vela_181.polarization * 100:.2f} %, Angle: {vela_181.angle:.2f} deg')
+    """
+
+    def __init__(self, pol0, pol60, pol120, filter_name, detector, pav3):
+        """
+        Parameters
+        ----------
+        pol0 : float
+            Photometric measurement in POL0 filter. Units: electrons or electrons/second.
```
pllim, line 59: Why not use `astropy.units.Quantity`? Same question for the other parameters with assumed units.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+        pol60 : float
+            Photometric measurement in POL60 filter. Units: electrons or electrons/second.
+        pol120 : float
+            Photometric measurement in POL120 filter. Units: electrons or electrons/second.
+        filter_name : str
+            Name of the filter crossed with the polarization filter, e.g., F606W.
+        detector : str
```
pllim, line 66:
```suggestion
        detector : {'wfc', 'hrc'}
```

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+
+        self.stokes_i = None
+        self.stokes_q = None
+        self.stokes_u = None
+
+        self.polarization = None
+        self.angle = None
+
+        # Get correction terms that we need from the polarization tables.
+        tables = PolarizerTables()
```
pllim, line 93: This needs to be more flexible in allowing different ways to provide input data (see review above on `PolarizerTables`). Perhaps `tables` itself can be an input to `Polarization`.
```suggestion
        tables = PolarizerTables()
```

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+
+class Polarization:
+    """
+    Class for handling ACS polarization data.
+
+    Examples
+    --------
+
+    >>> from acstools.polarization_tools import Polarization
+    >>> vela_181 = Polarization(63684, 67420, 63752, 'F606W', 'WFC', 348.084)
```
pllim, line 47: If you accept all my recommendations regarding API changes, it would be more like this:
```suggestion
    >>> from acstools.polarization_tools import Polarization, PolarizerTables
    >>> tables = PolarizerTables.from_package_data()
    >>> vela_181 = Polarization(tables, 63684, 67420, 63752, 'F606W', 'WFC', 348.084)
```

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+        self.stokes_i = None
+        self.stokes_q = None
+        self.stokes_u = None
+
+        self.polarization = None
+        self.angle = None
+
+        # Get correction terms that we need from the polarization tables.
+        tables = PolarizerTables()
+        leak_tab = Table(tables.table_data['transmission'][detector.lower()])
```
pllim, line 94: No need to call the `.lower()` and `.upper()` so many times. You can store the results in variables and use the latter.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+        self.transmission_correction = leak_tab[np.where(leak_tab['filter'] == filter_name.upper())]['correction'][0]
+
+        eff_tab = Table(tables.table_data['efficiency'][detector.lower()])
+        self.c0 = eff_tab[np.where(eff_tab['filter'] == filter_name.upper())]['pol0'][0]
```
pllim, line 98: Should there be a check that `==` actually returns a match?

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+        ----------
+        pol_measurements : list or tuple
+            Measurements in each of the three ACS polarization filters. Must be in order
+            of [POL0, POL60, POL120]. Units: electrons or electrons/second.
+        correction_terms : list or tuple
+            Polarizer efficiency correction values. Must be in order of [POL0, POL60, POL120].
+            These can be found in the tables provided by `acstools.acstools.polarization_tools.PolarizerTables()`
+            or in Table 5.6 in Section 5.3 of the ACS Data Handbook.
+
+        Returns
```
pllim, line 124: Hidden method's docstring is not rendered, so you need to copy the "Returns" info here to `calc_stokes` also if you decide to keep this as a separate hidden method.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+        self.c0 = eff_tab[np.where(eff_tab['filter'] == filter_name.upper())]['pol0'][0]
+        self.c60 = eff_tab[np.where(eff_tab['filter'] == filter_name.upper())]['pol60'][0]
+        self.c120 = eff_tab[np.where(eff_tab['filter'] == filter_name.upper())]['pol120'][0]
+
+    def calc_stokes(self):
+        """
+        Calculate Stokes parameters using attributes set at initialization.
+        """
+
+        self.stokes_i, self.stokes_q, self.stokes_u = self._calc_stokes([self.pol0, self.pol60, self.pol120],
```
pllim, line 107: Is there a reason why this calls a hidden method that isn't used anywhere else, instead of moving the code from `_calc_stokes` here and get rid of the hidden method?

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+
+        return i, q, u
+
+    def calc_polarization(self):
+        """
+        Calculate the polarization parameters (fractional polarization and position angle) for
+        using attributes set at initialization.
+        """
+
+        self.polarization = self._calc_fraction(self.stokes_i, self.stokes_q, self.stokes_u,
```
pllim, line 146: Similar comments about hidden methods and docstring as `calc_stokes`.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    Class for handling ACS polarization data.
+
+    Examples
+    --------
+
+    >>> from acstools.polarization_tools import Polarization
+    >>> vela_181 = Polarization(63684, 67420, 63752, 'F606W', 'WFC', 348.084)
+    >>> vela_181.calc_stokes()
+    >>> vela_181.calc_polarization()
```
pllim, line 49: A class makes sense if the methods can be called again and again with different inputs without having to go through the constructor again. But since your methods take no arguments, I wonder if a class is really necessary here. Should `Polarization` be re-designed into a function instead, which returns output `dict` mapping the polarization property names to values?

pllim (changes requested):
This is in addition to what I already mentioned above.

* Also need user facing documentation. If you already have that somewhere else (e.g., notebook or DHB), please link that to the user doc here. You can refer to #109 for example on how to do this as well.
* The file format that `PolarizerTables` expects should be documented clearly. What are the columns? What are the data types? What do they mean?
* Is your YAML file empty? GitHub seems to think it is.

Nice to do but at your discretion:

* Have `PolarizerTables` handle invalid input table gracefully. Currently, looks like it is going to throw some very uninformative `KeyError` or something if the wrong file is passed in.

acstools/polarization_tools.py
tddesjardins: Input values can be counts or rate, as long as they're consistent. And the values they transform into do not have units that transform logically. For example, Stokes I is unitless, but is the sum of three values that are either in counts or rate. So I think it's better to not accept `astropy.units.Quantity` objects for now, but I do think it's good to output the position angle as a quantity object to be clear it is in degrees.

acstools/polarization_tools.py
tddesjardins: I'm going to stick with the class approach, as the class is mostly a convenience to keep all the information packaged together in attributes, as well as to automatically grab the correct efficiencies and leak corrections from the table data. But I have broken out the hidden methods to independent functions so that users can use those instead of the class.

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+    For ACS, it is also necessary to supply the position angle of the V3 axis of the
+    telescope (PA_V3 in the primary header) and the name of the ACS detector used.
+
+    Parameters
+    ----------
+    q : float
+        Stokes Q parameter.
+    u : float
+        Stokes U parameter.
+    detector : ['wfc', 'hrc']
```
pllim, line 97: Curly bracket indicate the choices are in a set. That is the convention I learned anyway from SSB days.
```suggestion
    detector : {'wfc', 'hrc'}
```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+        Name of the ACS detector used for the observation. Must be either WFC or HRC.
+    pav3 : float
+        Position angle of the V3 axis in units of degrees. Found in ACS primary headers
+        with keyword PA_V3.
+
+    Returns
+    -------
+    theta : float
```
pllim, line 105: Why not return Quantity here?

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+
+    Parameters
+    ----------
+    q : float
+        Stokes Q parameter.
+    u : float
+        Stokes U parameter.
+    detector : ['wfc', 'hrc']
+        Name of the ACS detector used for the observation. Must be either WFC or HRC.
+    pav3 : float
```
pllim, line 99: Quantity?

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
 
-import numpy as np
 import os
+import numpy as np
 import yaml
 
+from astropy import units
 from astropy.table import Table
+from astropy.utils.data import get_pkg_data_filename
+
```
pllim, line 15: This is for star import support. Also used for API doc generation. Please fill in the rest and remove `...`.
```suggestion

__all__ = ['calc_stokes', 'calc_fraction', ...]

```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+    wfc_efficiency : Efficiencies of the ACS/WFC polarizers for computing Stokes parameters.
+    hrc_efficiency: Efficiencies of the ACS/HRC polarizers for computing Stokes parameters.
+
+    .. note::
+        The default table contained within the acstools package uses average transmission leak correction terms
+        for a source with a spectrum flat in wavelength space.
+
+    Polarizer calibration information can be read from the default YAML file contained in the
+    acstools package, or from a user-supplied YAML file using the class methods .from_yaml() and
```
pllim, line 135:
```suggestion
    acstools package, or from a user-supplied YAML file using the class methods :meth:`from_yaml` and
```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+    .from_package_data(). The YAML file format is:
```
pllim, line 136:
```suggestion
    :meth:`from_package_data`. The YAML file format is::
```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+                correction: list of transmission leak correction factors for each filter
+        efficiency:
+            detector:
+                filter: list of ACS filters
+                pol0: list of POL0 coefficients matching each filter
+                pol60: list of POL60 coefficients matching each filter
+                pol120: list of POL120 coefficients matching each filter
+
+    Multiple detectors can be contained in a single YAML file. An example is shown
+    below:
```
pllim, line 152:
```suggestion
    below::
```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+        Read in a YAML file containing polarizer calibration data.
+
+        Parameters
+        ----------
+        yaml_file : str
+            Path to the YAML file containing the polarizer calibration information.
+
+        Returns
+        -------
+        acstools.polarizer_tools.PolarizerTables object
```
pllim, line 233:
```suggestion
        pol_tables : `~acstools.polarizer_tools.PolarizerTables`
```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+
+    @classmethod
+    def from_package_data(cls):
+        """
+        Use the YAML file contained within the acstools package to retrieve the polarizer
+        calibration data.
+
+        Returns
+        -------
+        acstools.polarizer_tools.PolarizerTables object
```
pllim, line 247:
```suggestion
        pol_tables : `~acstools.polarizer_tools.PolarizerTables`
```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+    ----------
+    pol0 : float
+        Photometric measurement in POL0 filter. Units: electrons or electrons/second.
+    pol60 : float
+        Photometric measurement in POL60 filter. Units: electrons or electrons/second.
+    pol120 : float
+        Photometric measurement in POL120 filter. Units: electrons or electrons/second.
+    filter_name : str
+        Name of the filter crossed with the polarization filter, e.g., F606W.
+    detector : ['wfc', 'hrc']
```
pllim, line 270:
```suggestion
    detector : {'wfc', 'hrc'}
```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+        Name of the ACS detector used for the observation. Must be either WFC or HRC.
+    pav3 : float
```
pllim, line 272: Quantity?

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
+        Position angle of the HST V3 axis. This is stored in the ACS primary header under
+        keyword PA_V3. Units: degrees.
+    tables : acstools.polarization_tools.PolarizerTables
```
pllim, line 275:
```suggestion
    tables : `~acstools.polarization_tools.PolarizerTables`
```

acstools/polarization_tools.py
```diff
@@ -1,85 +1,328 @@
...
     >>> vela_181.calc_stokes()
     >>> vela_181.calc_polarization()
     >>> print(f'I = {vela_181.stokes_i:.2f}, Q = {vela_181.stokes_q:.2f}, U = {vela_181.stokes_u:.2f}')
-    >>> print(f'Polarization: {vela_181.polarization * 100:.2f} %, Angle: {vela_181.angle:.2f} deg')
+    I = 173336.09, Q = -3758.34, U = 9539.59
+    >>> print(f'Polarization: {vela_181.polarization:.2%}, Angle: {vela_181.angle:.2f}')
+    Polarization: 5.92%, Angle: 5.64 deg
+
+    If we need to adjust the polarization calibration, we can do so by providing a
+    different set of polarization tables using the acstools.polarization_tools.PolarizerTables
```
pllim, line 299:
```suggestion
    different set of polarization tables using the `~acstools.polarization_tools.PolarizerTables`
```

acstools/polarization_tools.py
```diff
@@ -90,116 +333,49 @@ def __init__(self, pol0, pol60, pol120, filter_name, detector, pav3):
...
-        """
-
-        chi = -38.2 if self.detector.lower() == 'wfc' else -69.4
-        theta = 0.5 * np.degrees(np.arctan2(u, q)) + pav3 + chi
-        theta -= 360 if theta > 360 else 0
+        self.polarization = calc_fraction(self.stokes_i, self.stokes_q, self.stokes_u,
+                                          transmission_correction=self.transmission_correction)
 
-        return theta
+        self.angle = calc_theta(self.stokes_q, self.stokes_u, self.detector, self.pav3) * units.degree
```
pllim, line 381: If `calc_theta` returns Quantity, manually applying `units.degree` here is no longer necessary.

pllim (commented):
Still missing:

* User facing (API) doc.
* Unit tests.

Need to add this section to `setup.cfg`, preferably after `[options.extra_require]`:

```
[options.package_data]
acstools = data/*
```

acstools/polarization_tools.py
tddesjardins: `quantity_input` doesn't quite work the way I'd like (it only checks if something is a quantity object and if not throws an exception), so I'm going to sneakily accept a quantity here but strip off the units (after ensuring it's in degrees) and then return a quantity at the end. The math in the middle just gets messy with having to throw in units on everything because it won't let me nicely add scalars to quantities.

acstools/polarization_tools.py
pllim: You need to fix this one to get Sphinx to generate API properly. That would fix linkcheck. The `__all__` need to include only the functions and classes you added in `polarization_tool.py`.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,414 @@
...
+
+    Parameters
+    ----------
+    q : float
+        Stokes Q parameter.
+    u : float
+        Stokes U parameter.
+    detector : {'wfc', 'hrc'}
+        Name of the ACS detector used for the observation. Must be either WFC or HRC.
+    pav3 : float
```
pllim, line 101:
```suggestion
    pav3 : float or `~astropy.units.Quantity`
```

acstools/polarization_tools.py
```diff
@@ -0,0 +1,414 @@
...
+    -------
+    theta : astropy.units.quantity.Quantity
+        Position angle of the electric field vector in degrees.
+    """
+
+    # If the user supplied an astropy Quantity object, strip off the units.
+    # This will make subsequent math easier without attaching units to
+    # everything. Make sure it's in degrees first.
+    if type(pav3) is units.quantity.Quantity:
+        pav3 = pav3.to(units.degree).value
```
pllim, line 115:
```suggestion
    if isinstance(pav3, units.quantity.Quantity):
        pav3 = pav3.to_value(units.degree)
```

acstools/polarization_tools.py
```diff
@@ -0,0 +1,414 @@
...
+
+    if detector.lower() not in ['wfc', 'hrc']:
+        raise ValueError('Detector must be either WFC or HRC.')
+
+    chi = -38.2 if detector.lower() == 'wfc' else -69.4
```
pllim, line 120: Please explain the magic numbers.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,414 @@
...
+        Stokes U parameter.
+    detector : {'wfc', 'hrc'}
+        Name of the ACS detector used for the observation. Must be either WFC or HRC.
+    pav3 : float
+        Position angle of the V3 axis in units of degrees. Found in ACS primary headers
+        with keyword PA_V3.
+
+    Returns
+    -------
+    theta : astropy.units.quantity.Quantity
```
pllim, line 107:
```suggestion
    theta : `~astropy.units.Quantity`
```

acstools/polarization_tools.py
pllim: Still need this in for proper API linking.

acstools/polarization_tools.py
pllim: Still need this in for proper API linking.

acstools/polarization_tools.py
```diff
@@ -0,0 +1,414 @@
...
+        Read in a YAML file containing polarizer calibration data.
+
+        Parameters
+        ----------
+        yaml_file : str
+            Path to the YAML file containing the polarizer calibration information.
+
+        Returns
+        -------
+        `~acstools.polarizer_tools.PolarizerTables` object
```
pllim, line 262: Need to reformat a bit for the API doc to show this part properly.
```suggestion
        pol_tables : `~acstools.polarizer_tools.PolarizerTables`
```

acstools/polarization_tools.py
```diff
@@ -0,0 +1,414 @@
...
+
+    @classmethod
+    def from_package_data(cls):
+        """
+        Use the YAML file contained within the acstools package to retrieve the polarizer
+        calibration data.
+
+        Returns
+        -------
+        `~acstools.polarizer_tools.PolarizerTables` object
```
pllim, line 276: Need to reformat a bit for the API doc to show this part properly.
```suggestion
        pol_tables : `~acstools.polarizer_tools.PolarizerTables` object
```

acstools/polarization_tools.py
```diff
@@ -0,0 +1,414 @@
...
+        Photometric measurement in POL0 filter. Units: electrons or electrons/second.
+    pol60 : float
+        Photometric measurement in POL60 filter. Units: electrons or electrons/second.
+    pol120 : float
+        Photometric measurement in POL120 filter. Units: electrons or electrons/second.
+    filter_name : str
+        Name of the filter crossed with the polarization filter, e.g., F606W.
+    detector : {'wfc', 'hrc'}
+        Name of the ACS detector used for the observation. Must be either WFC or HRC.
+    pav3 : float
```
pllim, line 301: Make it slightly less sneaky? 😸 
```suggestion
    pav3 : float or `~astropy.units.Quantity`
```

acstools/tests/test_polarization.py
```diff
@@ -0,0 +1,57 @@
+import pytest
+from .. import polarization_tools
+from astropy import units
```
pllim, line 3: I would swap these. Usual order is: stdlib, third-party, local
```suggestion
from astropy import units
from .. import polarization_tools
```

acstools/tests/test_polarization.py
```diff
@@ -0,0 +1,57 @@
...
+    """Test that the tables are loaded correctly and are all present."""
+    # Check if we can load the tables from the package data.
+    tables = polarization_tools.PolarizerTables.from_package_data()
+    assert hasattr(tables, 'wfc_transmission')
+    assert hasattr(tables, 'hrc_transmission')
+    assert hasattr(tables, 'wfc_efficiency')
+    assert hasattr(tables, 'hrc_efficiency')
+
+    # Check if we can load the tables by specifying the path to a YAML file.
+    _ = polarization_tools.PolarizerTables.from_yaml('acstools/data/polarizer_tables.yaml')
```
pllim, line 16: This is redundant and can be removed.
```suggestion
```

acstools/tests/test_polarization.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_theta():
+    """Test that the calc_theta function returns a quantity object
+    in degrees and that the value is as expected."""
+    # Check the angle for when we have all U polarization.
+    all_u = polarization_tools.calc_theta(0, 1, 'wfc', 38.2)
+    assert type(all_u) is units.quantity.Quantity
+    assert all_u.unit is units.degree
+    assert all_u.value == 45
```
pllim, line 26: Can be simplified by using [astropy.tests.helper.assert_quantity_allclose()](https://docs.astropy.org/en/stable/api/astropy.tests.helper.assert_quantity_allclose.html). If you want strict check, you can set `rtol` and `atol` accordingly.

```suggestion
    assert_quantity_allclose(all_u, 45 * units.degree)
```

acstools/tests/test_polarization.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # Also check the value for the all Q case.
+    all_q = polarization_tools.calc_theta(1, 0, 'wfc', 38.2)
+    assert all_q.value == 0
```
pllim, line 30:
```suggestion
    assert_quantity_allclose(all_q, 0 * units.degree)
```

acstools/tests/test_polarization.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_fraction():
+    """Test that the polarization fraction value is as expected."""
+
+    fraction = polarization_tools.calc_fraction(1, 0.353553, 0.353553)
+    assert fraction == pytest.approx(0.5, abs=1e-5)
```
pllim, line 37: This is more standard way to do this.
```suggestion
    np.testing.assert_allclose(fraction, 0.5, atol=1e-5)
```

acstools/tests/test_polarization.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_stokes():
+    """Test that the Stokes computation is correct."""
+
+    i, q, u = polarization_tools.calc_stokes(1, 1, 1, c0=1, c60=0.5, c120=1.5)
+    assert i == 2.0
+    assert q == 0.0
+    assert u == pytest.approx(-1.1547005383792517)
```
pllim, line 46:
```suggestion
    np.testing.assert_allclose(u, -1.1547005383792517)
```

pllim (commented):
Very nice!
```
acstools/tests/test_polarization.py::test_tables PASSED                  [  9%]
acstools/tests/test_polarization.py::test_theta PASSED                   [ 18%]
acstools/tests/test_polarization.py::test_fraction PASSED                [ 27%]
acstools/tests/test_polarization.py::test_stokes PASSED                  [ 36%]
acstools/tests/test_polarization.py::test_pol_class PASSED               [ 45%]
```

Some more comments, but we're very close! Thanks for your continued patience with this review process.

pllim (approved):
🐴 💀 :shipit: 

Jenkins failures are unrelated. Let's merge this. Thanks!

Please let me know if you want a release now, or you want to wait a bit.

tddesjardins:
The release can wait a bit. I'd like to go over some of this with Dean first just to make sure, and then we'll add the text to the handbooks. Thanks!

PR #135, opened 2020-09-23
MNT: New way to build sdist

pllim:
Redo of f556c6f , with fix suggested by @Cadair .


pllim:
Jenkins actually failed but the failure is unrelated anyway.

PR #136, opened 2020-10-08
TST: Avoid error when pytest-astropy-header not installed

pllim:
Header won't show but at least it won't crash.


rendinam (approved):
Looks good to me! Thanks!

PR #140, opened 2020-11-10
Updated handling of extrapolated dates

### LINKED ISSUE #137: Remove upper limit on input date in acszpt
The `acszpt` module requires that the input date of the calculation be earlier than 12-31-2021: https://github.com/spacetelescope/acstools/blob/62c02b3df3534a89a070d450d6759bc187cdd07c/acstools/acszpt.py#L168-L170

The ACS zeropoints tool webapp does not throw an error or warning for the middle of Cycle 29 (04-01-2022): https://acszeropoints.stsci.edu/results_all/?date=2022-04-01&detector=WFC

There is a date upper limit (in MJD) built into the throughput tables, so perhaps the `acszpt` module or the zeropoints tool could read that date from those files?

molaes-stsci:
- fixed minor typo in example
- removed usage of locally defined "extrapolation_date"
- added functionality to pass warnings that have been generated from ZPT tool, if any
Fixes #137

### DIFF
--- a/acstools/acszpt.py
+++ b/acstools/acszpt.py
@@ -49,7 +49,7 @@ Retrieve the zeropoint information for the F435W filter on 2016-04-01 for WFC:
 >>> date = '2016-04-01'
 >>> detector = 'WFC'
 >>> filt = 'F435W'
->>> q = acszpt.Query(date=date, detector=detector, filter=filt)
+>>> q = acszpt.Query(date=date, detector=detector, filt=filt)
 >>> zpt_table = q.fetch()
 >>> print(zpt_table)
 FILTER PHOTPLAM        PHOTFLAM         STmag  VEGAmag  ABmag
--- a/acstools/acszpt.py
+++ b/acstools/acszpt.py
@@ -375,8 +375,8 @@ class Query:
 
             LOG.info('Parsing the response and formatting the results...')
             self._parse_and_format()
-            if self._warnings:
-                _ = [LOG.warning(w) for w in self._warnings]
+            for w in self._warnings:
+                LOG.warning(w)
 
             return self.zpt_table
 
--- a/acstools/acszpt.py
+++ b/acstools/acszpt.py
@@ -153,6 +153,7 @@ class Query:
                     'F140LP', 'F150LP', 'F165LP']
         }
         self._zpt_table = None
+        self._warnings = None
 
         # Set the private attributes
         if filt is None:
@@ -165,9 +166,6 @@ class Query:
                          f'&{self.detector}_filter={self.filt}')
         # ACS Launch Date
         self._acs_installation_date = dt.datetime(2002, 3, 7)
-        # The farthest date in future that the component and throughput files
-        # are valid for. If input date is larger, extrapolation is not valid.
-        self._extrapolation_date = dt.datetime(2021, 12, 31)
         self._msg_div = '-' * 79
         self._valid_detectors = ['HRC', 'SBC', 'WFC']
         self._response = None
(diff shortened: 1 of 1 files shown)


acstools/acszpt.py
```diff
@@ -49,7 +49,7 @@
 >>> date = '2016-04-01'
 >>> detector = 'WFC'
 >>> filt = 'F435W'
->>> q = acszpt.Query(date=date, detector=detector, filter=filt)
+>>> q = acszpt.Query(date=date, detector=detector, filt=filt)
```
pllim, line 52: Good catch!

acstools/acszpt.py
```diff
@@ -319,6 +310,15 @@ def _parse_and_format(self):
...
         td = [val.text.split(' ')[0] for val in td]
 
+        # Grab warnings returned by the ZPT calc as marked by stylized h4 tag
+        warnings = soup.find_all('h4', {'style': "color:red;"})
+
+        # Remove tags and 'Warnings: ' label attached to warnings
+        str_warnings = [w.get_text().replace('Warnings: ', '') for w in warnings]
+
+        # Add non-empty strings to _warnings attribute
+        self._warnings = [w for w in str_warnings if w.strip()]
```
pllim, line 320: I feel like this all can be collapsed into one line. But I don't understand what kind of warnings the server is going to issue to give a recommendation.

acstools/acszpt.py
```diff
@@ -153,6 +153,7 @@ def __init__(self, date, detector, filt=None):
                     'F140LP', 'F150LP', 'F165LP']
         }
         self._zpt_table = None
+        self._warnings = None
```
pllim, line 156: Probably easier if you keep it as a list throughout.
```suggestion
        self._warnings = []
```

acstools/acszpt.py
```diff
@@ -375,6 +375,9 @@ def fetch(self):
 
             LOG.info('Parsing the response and formatting the results...')
             self._parse_and_format()
+            if self._warnings:
+                _ = [LOG.warning(w) for w in self._warnings]
```
pllim, line 379: No need to construct a list.
```suggestion
            for w in self._warnings:
                LOG.warning(w)
```

pllim (commented):
Given that this module has no unit tests in the CI (probably because it hits AWS and costs us money), I need someone in ACS Team to test that this works, especially for the extrapolation case, and approve. @jryon ?

Also some minor comments below.

In the longer term, please consider to have the server return a JSON file in addition to HTML. That would make parsing logic way easier.

acstools/acszpt.py
molaes-stsci: It can be, but it just looked messy as one line. I have it this way just in case there are multiple warning lines passed, but I'm not well versed in the possible output of the ZPT tool

acstools/acszpt.py
molaes-stsci: Thanks! This is a much better solution.

acstools/acszpt.py
pllim: Okay, let's leave it as-is then. 😸

acstools/acszpt.py
pllim: Hmm. Did you forget to commit this suggestion?

molaes-stsci: Why yes, yes I did. Committed! Thanks.

pllim:
LGTM but I would like a second pairs of eyes from ACS Team to look at this before approving.

jryon:
I'll take a look, thanks!

jryon (approved):
Looks good to me. Behaves as expected given webtool warnings. No longer fails when date gets arbitrarily large.

Uncovered a minor issue with the warnings passed to this module by the webtool, but shouldn't stop this PR from being approved.

pllim:
> Uncovered a minor issue

Can you please elaborate on what the issue is? Should that issue be addressed as part of this PR or separately?

jryon:
The issue is in the extrapolation checking from the zeropoints webtool, which is completely separate software from acstools.  It can't be addressed by this PR.

pllim (approved):
Good to know, merging then. Thanks, everyone!

I am still waiting for a "go" from @tddesjardins for the next `acstools` release. So this patch will not be released until that happens. FYI.

PR #157, opened 2021-12-21
TST: Fix CI failures

pllim:
### Description

This pull request is to address warnings and remote data failures due to upstream changes. All the deprecation warnings are from upstream packages:

* `astropy`: Fixed in stable version that requires Python 3.8 or newer, so jobs using Python 3.7 won't see that fix.
* `pytest-remotedata`: astropy/pytest-remotedata#58
* `stsci.tools`: spacetelescope/stsci.tools#144


pllim:
No RTD failure makes no sense. I will merge anyway and see if clearing cache on RTD will fix things. I cannot clear cache for PRs.

PR #169, opened 2022-09-14
updated photcte.py with new api id and key

gsanand:
### Description
This pull request is to update the photcte.py module with the updated API ID and Key for the new AWS control tower environment deployment requested by ITSD.


jryon:
Looks good to me. Does the "Draft" status of this PR mean you don't want it merged now?

gsanand:
I think my initial thought was to not have it updated until ITSD is done on their end but since it's not being pushed to pip until after that anyways it doesn't matter, so I'll push it past draft now

jryon (approved):
Looks good!

PR #170, opened 2022-09-20
use canonical filename of RTD configuration and enable `nitpicky` mode

zacharyburnett:
### Description

update the name of the ReadTheDocs configuration file to `.readthedocs.yaml` and add `nitpicky` to validate links


pllim:
Ooof, 176 warnings. Can we not do nitpicky right now?

zacharyburnett:
> Ooof, 176 warnings. Can we not do nitpicky right now?

yes, I think that is enough for another PR

pllim (approved):
Thanks!

PR #173, opened 2022-12-07
Include Zenodo DOI for v3.5.0 in README

jryon:
### Description

This pull request is to update the README with the new Zenodo DOI.


pllim (changes requested):
Just having 2 badges listed without any explanation might be too confusing. If you want to keep both, you have to explain how each should be used.

Also please update https://github.com/spacetelescope/acstools/blob/master/CITATION.md

pllim (approved):
I took the liberty to push a commit to README. LGTM now if you okay with that. Thanks!

jryon:
Works for me. Thanks!

PR #180, opened 2023-09-28
added first iteration of focus_diverse_epsfs

gsanand:
added first iteration of a new module called  focus_diverse_epsfs

### DIFF
--- a/acstools/focus_diverse_epsfs.py
+++ b/acstools/focus_diverse_epsfs.py
@@ -27,36 +27,36 @@ Retrieve a single 101x101x90 ePSF FITS file.
 
 Retrieve ePSFs based on flc images specified in an external text file (with one rootname per line).
 
->>>input_list = 'input_ipsoots.txt'
+>>> input_list = 'input_ipsoots.txt'
 >>> download_location  = '/Users/username/download_folder/'
->>>retrieved_downloads = multi_psf_retriever(input_list, download_location, fromTextFile = True)
+>>> retrieved_downloads = multi_psf_retriever(input_list, download_location, fromTextFile = True)
 
 Retrieve ePSFs using rootnames obtained from an astroquery.
 
 >>> download_location  = '/Users/username/download_folder/'
->>>obsTable = Observations.query_criteria(obs_collection = 'HST', proposal_id="13376", instrument_name = "ACS/WFC",
+>>> obsTable = Observations.query_criteria(obs_collection = 'HST', proposal_id="13376", instrument_name = "ACS/WFC",
                                        provenance_name = "CALACS")
 
->>>dataProducts = Observations.get_product_list(obsTable)
->>>dataProducts = dataProducts[(dataProducts['productSubGroupDescription'] == 'FLC') &
+>>> dataProducts = Observations.get_product_list(obsTable)
+>>> dataProducts = dataProducts[(dataProducts['productSubGroupDescription'] == 'FLC') &
              (dataProducts['type'] == 'S')]
 
->>>obs_rootnames = list(dataProducts['obs_id'])
->>>retrieved_downloads = multi_psf_retriever(obs_rootnames, download_location)
+>>> obs_rootnames = list(dataProducts['obs_id'])
+>>> retrieved_downloads = multi_psf_retriever(obs_rootnames, download_location)
 
 
 Interpolate a given ePSF to x,y = (2000,4048), which is near the middle of the detector along the x-axis,
 and near the top of the WFC1 chip (and the detector overall).
 
->>>x = 2000
->>>y = 4048
->>>P = interp_epsf(ePSFs, x, y)
+>>> x = 2000
+>>> y = 4048
+>>> P = interp_epsf(ePSFs, x, y)
 
 Do the same as the previous example, but obtain the ePSF in detector space (instead of with 4x supersampling).
->>>P = interp_epsf(ePSFs, x, y, pixel_space = True)
+>>> P = interp_epsf(ePSFs, x, y, pixel_space = True)
 
 Do the same as the previous example, but now in detector space and with subpixel offsets.
->>>P = interp_epsf(ePSFs, x, y, pixel_space = True, subpixel_x = 0.77, subpixel_y = 0.33)
+>>> P = interp_epsf(ePSFs, x, y, pixel_space = True, subpixel_x = 0.77, subpixel_y = 0.33)
 
 
 More details for these examples are provided in a Jupyter notebook, which can be found at LINK.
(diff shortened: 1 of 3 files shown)


pllim (commented):
What is the deadline for this, so I can prioritize among my other work?

Some general comments from initial quick glance:

* You need to hook up the API doc to package documentation, if it is not getting picked up already.
* Avoid referring to your own directory in docs. Use things others can access like `astroquery.mast` calls if the file is available from MAST.
* Do you plan to add unit tests?
* Is that gateway URL static? What about the auth info? Should it be public like that (is it guarded against DDoS attacks)? Does it take additional auth tokens that are private? Do not put anything private here.
* Are you handling network issues gracefully? What if your query returns a HTTP Error 500 and so on?
* Is it appropriate to use `print` instead of throwing proper exception in a function?
* Do you really have to roll your own interpolation algorithm or can you use an existing one from photutils, scikit-image, scipy, numpy, and so on?
* dask is an undeclared dependency
* scipy is only an optional dependency
* Please arrange imports in the order of stdlib, blankline, third-party (and each block, do the `import` first before `from`, and then alphabetically)

gsanand:
Hi @pllim, in terms of deadline, we don't have a strict timeline, but ~2 to 3 weeks out from now is fine.

I'll take a look at addressing some of your comments. regarding the DDoS protection/auth info, that is the public API ID and Key we have generated, and we have quotas in place to prevent excessive usage on the AWS API Gateway end of things.

acstools/focus_diverse_epsfs.py
```diff
@@ -0,0 +1,349 @@
...
+
+    Downloads the ePSF FITS files to download_location. The explicit return is 
+    a list of the downloaded FITS files.
+
+    """
+
+    # select if either choosing to provide inputs via a text file or list
+    if fromTextFile == True:
+        with open(input_list) as file:
+            lines = [line.rstrip() for line in file]
```
jryon, line 147: This is something @pllim may have more to say about, but I think you'll need some smart handling/errors for poorly-formatted input files.

acstools/focus_diverse_epsfs.py
```diff
@@ -0,0 +1,349 @@
...
+    Function to query API on AWS API Gateway for the ePSF FITS file that
+    corresponds to a given image rootname.
+
+    Returns
+    -------
+
+    Downloads the ePSF FITS file to download_location. The explicit return is 
+    the name of the downloaded FITS file.
+
+    """
```
jryon, line 100: Check docstrings of other acstools modules for preferred format/content.

acstools/focus_diverse_epsfs.py
```diff
@@ -0,0 +1,349 @@
...
+    subpixel_y = round(subpixel_y, 2)
+
+    # if user inputs coordinates for WFC1 (y > 2048), then adjust down and set chip = "WFC1",
+    # otherwise assume chip = "WFC2"
+    chip = "WFC2"
+
+    if y > 2048:
+        y = y-2048
+        chip = "WFC1"
+
```
jryon, line 226: Make it very clear in the docstring/docs that the expected coordinates here are for the whole 4096x4096 frame, not per chip. I'm not 100% sure I agree with this choice, since I'd imagine most people will be choosing their desired PSF location in an FLT/FLC, which has separate arrays for each chip. Maybe there's a way to enable both full-frame and per chip coordinates?

jryon (commented):
Looking good so far. I'll test out functionality after I return from vacation (Oct 11).

pllim:
@gsanand , can you please rebase? I just overhauled the CI stuff. Thanks!

```
git fetch upstream master
git rebase upstream/master
```

You might have to resolve conflicts or not. Then when you are sure things went well:

```
git push -f origin master
```

jryon:
I did some basic functionality testing and things are working well. I have a few comments:

- The trailing forwardslash '/' appears to be relevant for the download location. I accidentally left it off at first and wondered where my download went. The filename had the intended directory as a prefix. I'd suggest you add a forwardslash to whatever directory the user inputs.
- The progress bar seems a little clunky. In one test, it behaved about as expected, but in another it stayed at 0% for several seconds, then went straight to 100%.
- Forgive me if this is discussed in your ISR and I forgot - are subarrays not supported, then, since this module requires an FLC? Could we use an FLT instead or Is there some other reason for not supporting subarrays?

acstools/focus_diverse_epsfs.py
```diff
@@ -0,0 +1,395 @@
...
+Additionally, this module also contains a function (interp_epsf) that allows
+users to interpolate the provided ePSF arrays to any arbitrary (x,y)
+coordinates. This function can be called with "pixel_space" = True to
+downsample the ePSF into detector space. Subpixel phase shifts can be applied by
+setting subpixel_x and subpixel_y between 0.00 and 0.99.
+
+Examples
+--------
+
```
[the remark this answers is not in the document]
gsanand: yes it was in the ISR, but I should have made the changes to the module as well. will do along with the next commit.

gsanand:
Hey @jryon, thanks for the comments!

- I implemented a fix for this so hopefully it should not be an issue anymore.

- yes the progress bar is not great sometimes (slow to update). I looked into it and it just seems to be an issue with the underlying code, which I can't fix. I can take out the progress bar if you think that might be the better option.

-subarrays are indeed not supported (because of the support only for FLCs). I actually did not state this in the ISR before, so I just added a line saying as much.

jryon:
> * yes the progress bar is not great sometimes (slow to update). I looked into it and it just seems to be an issue with the underlying code, which I can't fix. I can take out the progress bar if you think that might be the better option.

In this case, I'm fine with leaving it, since it is useful at least some of the time.

acstools/focus_diverse_epsfs.py
```diff
@@ -0,0 +1,406 @@
...
+Do the same as the previous example, but obtain the ePSF in detector space (instead of with 4x supersampling).
+>>> from focus_diverse_epsfs import interp_epsf
+>>> P = interp_epsf(ePSFs, x, y, pixel_space = True)
+
+Do the same as the previous example, but now in detector space and with subpixel offsets.
+>>> from focus_diverse_epsfs import interp_epsf
+>>> P = interp_epsf(ePSFs, x, y, pixel_space = True, subpixel_x = 0.77, subpixel_y = 0.33)
+
+
+More details for these examples are provided in a Jupyter notebook, which can be found at LINK.
```
pllim, line 71: I am going to delete this because this package isn't responsible for keeping any notebook elsewhere up-to-date (in fact, notebooks tend to quickly break and unmaintained). So you can mention it in your ISR but you do not have to mention it here.

pllim:
Sorry, I had to create #183 because of reasons... should we continue there?

pllim:
Closing in favor of #183. Thanks for your patience with me!

PR #185, opened 2023-11-07
TST: Fix CI failures in devdeps

### LINKED ISSUE #182: TST: test_wfc_findsat_mrt.py has MergeConflictWarning using astropy 6.0.dev
@dvstark , you should to look into why there is a warning and whether your data or test needs fixing or this warning is expected. Thanks.

You can also install dev version of astropy locally and debug this. See https://docs.astropy.org/en/latest/install.html#installing-pre-built-development-versions-of-astropy

Example log: https://github.com/spacetelescope/acstools/actions/runs/6543186852/job/17767366293

```
Package versions: 
Numpy: 2.0.0.dev0+git20231014.dbd30ce
Scipy: 1.12.0.dev0+1845.7819f1b
Matplotlib: 3.9.0.dev355+g21ce59212d
Astropy: 6.0.dev865+gd64fc33942
beautifulsoup4: not available
requests: 2.31.0
stsci.tools: 4.1.0
...
_______________________________ test_WfcWrapper ________________________________

left = OrderedDict({'date': '2023-10-17 06:18:23 UTC', 'versions': {'Python': '3.12.0', 'photutils': '1.9.1.dev99+g93e1[145](https://github.com/spacetelescope/acstools/actions/runs/6543186852/job/17767366293#step:10:146)2',...v0+git20231004.c01ae65', 'sklearn': None, 'matplotlib': '3.9.0.dev355+g21ce59212d', 'gwcs': None, 'bottleneck': None}})
right = OrderedDict({'date': '2023-10-17 06:18:24 UTC', 'versions': {'Python': '3.12.0', 'photutils': '1.9.1.dev99+g93e11452',...v0+git20231004.c01ae65', 'sklearn': None, 'matplotlib': '3.9.0.dev355+g21ce59212d', 'gwcs': None, 'bottleneck': None}})
merge_func = None, metadata_conflicts = 'warn'
warn_str_func = <function _warn_str_func at 0x7fa02e6e0d60>
error_str_func = <function _error_str_func at 0x7fa02e6e0e00>

    def merge(
        left,
        right,
        merge_func=None,
        metadata_conflicts="warn",
        warn_str_func=_warn_str_func,
        error_str_func=_error_str_func,
    ):
...
>                           raise MergeConflictError
E                           astropy.utils.metadata.exceptions.MergeConflictError

.../astropy/utils/metadata/merge.py:303: MergeConflictError

During handling of the above exception, another exception occurred:
...
left = OrderedDict({'date': '2023-10-17 06:18:23 UTC', 'versions': {'Python': '3.12.0', 'photutils': '1.9.1.dev99+g93e11452',...v0+git20231004.c01ae65', 'sklearn': None, 'matplotlib': '3.9.0.dev355+g21ce59212d', 'gwcs': None, 'bottleneck': None}})
right = OrderedDict({'date': '2023-10-17 06:18:24 UTC', 'versions': {'Python': '3.12.0', 'photutils': '1.9.1.dev99+g93e11452',...v0+git20231004.c01ae65', 'sklearn': None, 'matplotlib': '3.9.0.dev355+g21ce59212d', 'gwcs': None, 'bottleneck': None}})
merge_func = None, metadata_conflicts = 'warn'
warn_str_func = <function _warn_str_func at 0x7fa02e6e0d60>
error_str_func = <function _error_str_func at 0x7fa02e6e0e00>

    def merge(
        left,
        right,
        merge_func=None,
        metadata_conflicts="warn",
        warn_str_func=_warn_str_func,
        error_str_func=_error_str_func,
    ):
...
>                           warnings.warn(
                                warn_str_func(key, left[key], right[key]),
                                MergeConflictWarning,
E                               astropy.utils.metadata.exceptions.MergeConflictWarning: Cannot merge meta key 'date' types <class 'str'> and <class 'str'>, choosing date='2023-10-17 06:18:24 UTC'

.../astropy/utils/metadata/merge.py:321: MergeConflictWarning
...
------------------------------ Captured log call -------------------------------
INFO     findsat_mrt:findsat_mrt.py:1255 image type is flc
INFO     findsat_mrt:findsat_mrt.py:1322 masking bad pixels
INFO     findsat_mrt:findsat_mrt.py:1341 Subtracting median background
INFO     findsat_mrt:findsat_mrt.py:1361 Rebinning the data by 4
INFO     findsat_mrt:findsat_mrt.py:1283 Running the trailfinding pipeline
INFO     utils_findsat_mrt:utils_findsat_mrt.py:1083 Calculating median Radon Transform with 2 processes
INFO     findsat_mrt:findsat_mrt.py:763 Detection threshold: 5
INFO     findsat_mrt:findsat_mrt.py:773 Using kernel .../acstools/data/rt_line_kernel_width15.fits
INFO     findsat_mrt:findsat_mrt.py:789 {no} sources found using kernel: 3
INFO     findsat_mrt:findsat_mrt.py:793 3 sources found using kernel
INFO     findsat_mrt:findsat_mrt.py:773 Using kernel .../acstools/data/rt_line_kernel_width7.fits
INFO     findsat_mrt:findsat_mrt.py:789 {no} sources found using kernel: 7
INFO     findsat_mrt:findsat_mrt.py:793 7 sources found using kernel
INFO     findsat_mrt:findsat_mrt.py:773 Using kernel .../acstools/data/rt_line_kernel_width3.fits
INFO     findsat_mrt:findsat_mrt.py:789 {no} sources found using kernel: 24
INFO     findsat_mrt:findsat_mrt.py:793 24 sources found using kernel
INFO     findsat_mrt:findsat_mrt.py:803 Removing duplicate sources
```

### LINKED ISSUE #184: TST: 5 test failures in test_wfc_destripe.py in devdeps
Example log: https://github.com/spacetelescope/acstools/actions/runs/6627739837/job/18206557160

```
Package versions: 
Numpy: 2.0.0.dev0+git20231027.cdfbdf4
Scipy: 1.12.0.dev0+1899.e35850d
Matplotlib: 3.9.0.dev355+g21ce59212d
Astropy: 6.1.dev0
beautifulsoup4: not available
requests: 2.31.0
stsci.tools: 4.1.0
...
../../acstools/tests/test_wfc_destripe.py::TestDestripe::test_generic[jb5g05ubq] FAILED [ 50%]
../../acstools/tests/test_wfc_destripe.py::TestDestripe::test_generic[ja0x03ojq] FAILED [ 58%]
../../acstools/tests/test_wfc_destripe.py::TestDestripe::test_generic[jc5001soq] FAILED [ 66%]
../../acstools/tests/test_wfc_destripe.py::TestDestripe::test_fullframe_flashed FAILED [ 75%]
../../acstools/tests/test_wfc_destripe.py::TestDestripe::test_fullframe_masked FAILED [ 83%]
```

These appears to be floating point differences. So someone who is familiar with the destriping algorithm should double check to make sure that these differences are acceptable, and if not, what kind of fixes are needed to make destriping compatible with unreleased versions of numpy/scipy/astropy/etc.

```
E           Extension HDU 1 (SCI, 1):
E           
E              Data contains differences:
E                Data differs at [np.int64(727), np.int64(6)]:
E                   a> np.float32(0.9507208)
E                    ?                   ^^
E                   b> np.float32(0.9507245)
E                    ?    
```

pllim:
### Description

This pull request is to fix CI failures seen in devdeps job (unreleased upstream dependencies):

* fix #182 
* fix #184

### DIFF
--- a/acstools/tests/test_wfc_destripe.py
+++ b/acstools/tests/test_wfc_destripe.py
@@ -26,7 +26,7 @@ class TestDestripe(BaseACSTOOLS):
                                mask1=masks[0], mask2=masks[1])
 
         # Compare results
-        self.compare_outputs([(outputfile, truthfile)], rtol=1e-6, atol=0.1)
+        self.compare_outputs([(outputfile, truthfile)], rtol=1e-3)
 
     # jb5g05ubq = calibrated post-SM4 WFC full-frame exposures without masks
     # ja0x03ojq = calibrated polarizer WFC subarray exposures
--- a/acstools/tests/test_wfc_destripe.py
+++ b/acstools/tests/test_wfc_destripe.py
@@ -26,7 +26,7 @@ class TestDestripe(BaseACSTOOLS):
                                mask1=masks[0], mask2=masks[1])
 
         # Compare results
-        self.compare_outputs([(outputfile, truthfile)], rtol=1e-6)
+        self.compare_outputs([(outputfile, truthfile)], rtol=1e-6, atol=0.1)
 
     # jb5g05ubq = calibrated post-SM4 WFC full-frame exposures without masks
     # ja0x03ojq = calibrated polarizer WFC subarray exposures
--- a/acstools/tests/test_wfc_destripe.py
+++ b/acstools/tests/test_wfc_destripe.py
@@ -26,7 +26,7 @@ class TestDestripe(BaseACSTOOLS):
                                mask1=masks[0], mask2=masks[1])
 
         # Compare results
-        self.compare_outputs([(outputfile, truthfile)], rtol=1e-3)
+        self.compare_outputs([(outputfile, truthfile)], rtol=1e-2)
 
     # jb5g05ubq = calibrated post-SM4 WFC full-frame exposures without masks
     # ja0x03ojq = calibrated polarizer WFC subarray exposures
--- a/acstools/findsat_mrt.py
+++ b/acstools/findsat_mrt.py
@@ -1104,7 +1104,7 @@ class TrailFinder:
                 # an inconsequential warning so we suppress it here.
                 with warnings.catch_warnings():
                     warnings.filterwarnings('ignore',
-                                            message='Attribute `version` of')
+                                            message='Attribute `version.*` of')
                     self.source_list.write(cat_filename, overwrite=True)
             else:
                 # create an empty catalog and write that. It helps to have this
--- a/acstools/utils_findsat_mrt.py
+++ b/acstools/utils_findsat_mrt.py
@@ -108,7 +108,9 @@ def merge_tables(tbls, theta_sep=10, rho_sep=10):
             drho = np.abs(t['ycentroid'] - s['ycentroid'])
             sel = (dtheta < theta_sep) & (drho < rho_sep)
             keep[sel] = False
-        src = vstack([src, t[keep]])
+        # Silence MergeConflictWarning for "date" metadata because
+        # the tables were created with different timestamps. The last entry is kept.
+        src = vstack([src, t[keep]], metadata_conflicts="silent")
     return src
 
 


acstools/tests/test_wfc_destripe.py
```diff
@@ -26,7 +26,7 @@ def _destripe_one(self, inputfile, outsuffix, outputfile, truthfile,
                                mask1=masks[0], mask2=masks[1])
 
         # Compare results
-        self.compare_outputs([(outputfile, truthfile)], rtol=1e-6)
+        self.compare_outputs([(outputfile, truthfile)], rtol=1e-6, atol=0.1)
```
pllim, line 29: @jryon , is setting `atol=0.1` controversial? Anything lower would cause some comparison to fail. Do you care about disagreement smaller than this in the pixel values?

jryon: So including the atol causes these tests not to fail anyore? My issue with an absolute tolerance is that the SCI arrays in the files being tested cover a lot of dynamic range, from what I could see. Some have pixel values of order 0.05 electrons and others are 100 electrons. Why not increase rtol or redefine truth instead?

pllim: If we only want to touch rtol, it would have to be something like `rtol=1e-3` or larger. Here is one of the disagreements:

```
                    a> np.float32(21.261755)
E                    ?                ^ ---
E                   b> np.float32(21.25891)
E                    ?                ^^^
```

This disagreement only shows up using unreleased versions of upstream dependencies (probably because numpy 2.0.dev changed a bunch of stuff, but also could be scipy or astropy here, so not 100% sure). So if you update the truth, the other jobs using released versions would fail instead.

Another option is for someone on the ACS Team to spend time investigating why the results are different using unreleased numpy/scipy/astropy etc.

jryon: I don't see that disagreement in the previous error log you shared, and I don't know where to find the others - can you point me to them? This one is slightly larger than discrepancies I saw, so I'd like to take a look at all of the errors.

Forgive my naivety, but why are we testing against unreleased versions as well as released versions of dependencies? Why not just test against released versions and update truth/fix problems at that point?

dvstark (approved):
I'm good with the changes to findsat_mrt.py.

acstools/tests/test_wfc_destripe.py
pllim: > I don't see that disagreement in the previous error log you shared

The way fitsdiff works, you won't see everything. It only popped up for me when I started relaxing the tolerance bit by bit in a local run after I managed to set up the environment locally so that the tests fail in the same way as CI.

> why are we testing against unreleased versions as well as released versions of dependencies?

It is to future-proof the package. Both numpy 2.0 and astropy 6.0 (both major releases but numpy 2.0 is the major refactor one) are slated to be released by the end of the year (so in a month or two). To have a matrix of both released and unreleased versions of dependencies would ensure (1) your code is compatible with things we know are getting released in the future, and (2) any changes to fix 1 won't accidentally break backward compatiblity with already released versions. This is more or less a standard practice in larger libraries that I try to emulate here.

> Why not just test against released versions and update truth/fix problems at that point?

That is also an option you can consider. However, this means you need to bump the minimum required versions to the latest versions almost immediately after they are released. Given how breaking numpy 2.0 is going to be, not all your other dependencies can catch up quickly, so you might be stuck with a very broken CI for many months.

Can you also update the truth without bumping required minimum versions? Yes, you can, but it is a promise you cannot keep because you no longer is able to test against the minimum versions you are advertising due to truth updates.

Does this make sense?

jryon: @pllim Yes, this makes sense. Thanks for explaining. I certainly don't want to change what we're doing if it's essentially standard practice, and it sounds like a solid system. I'm happy to go with an rtol bump instead of atol, and I'll discuss again with Norman to make sure we're still ok with discrepancies slightly larger than the logs say.

pllim:
@jryon , as you can see, I relaxed rtol all the way to `1e-2` and there are still some failures though much less, and now they all seem to be very small numbers.

As mentioned above, there are many ways we can approach this, including just letting it fail until numpy/astropy/etc release new versions. I would let the ACS Team decide what to do here.

pllim:
I managed to get the new results out of the test folder. I will DM you about how to share them. Here is example visual difference for `ja0x03ojq_flt` test case:

ja0x03ojq_flt_diff_compare

jryon:
@pllim To clarify, if we were to ignore the failing tests for now, would we then redefine truth once the new versions of astropy/etc are released? Or would we still need to bump atol/rtol in order to keep compatibility with older versions of dependencies? Basically wondering if ignoring failures for now is just kicking the can down the road.

pllim:
> Basically wondering if ignoring failures for now is just kicking the can down the road.

Yes, it is. I can skip the tests for `numpy>=2` to make things green again for destripe but that means they will not be tested at all when numpy 2.0 is released for any job that picks up latest stable numpy automatically until we go back and figure it out later.

Like I said, there are a lot of options here. Just a matter of what ACS Team think is important and what can wait. Thanks!

pllim:
Since destripe needs more investigation, I will close this PR without merge and apply MRT fix separately. Please let me know when you have a decision on how you want to proceed with the destripe stuff. Thanks!

PR #190, opened 2023-11-13
BLD: Do not use OpenAstronomy workflow

pllim:
### Description

This pull request is to stop using OpenAstronomy workflow. Building simple wheel isn't that hard anyway. We can just copy paste from what I know works over at `stsynphot`.


pllim:
We do not want to bundle stuff in LFS. But there is also no point to test the wheel without the data. So you win some, you lose some?

No need to wait for CI. YOLO.

Issue #193, opened 2024-01-02
TST: acsphotometriccte linkcheck failed with [SSL: CERTIFICATE_VERIFY_FAILED]


pllim:
Example log: https://github.com/spacetelescope/acstools/actions/runs/7382744253/job/20082925111

`doc/source/acsphotcte.rst:1:broken link`

https://acsphotometriccte.stsci.edu/

`SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)'))))`

PR #194, opened 2024-01-02
Fix devdeps

pllim:
### Description

This pull request is to fix devdeps because photutils got installed after matplotlib downgraded numpy now.


pllim:
Link check failure is unrelated (https://github.com/spacetelescope/acstools/issues/193).

PR #196, opened 2024-02-05
satdet.detsat fix for scipy.stats.mode

dvstark:
### Description

This pull request is to address an error that can sometimes occur in acstools.satdet.detsat routine. This function calls scipy.stats.mode and is expecting an array as output. Historically, scipy.stats.mode always returned an array, but after v1.9.2 it returns a scalar in certain situations. Setting the keepdims keyword to True ensures the output is an array, ensuring compatibility acstools.

### DIFF
--- a/acstools/satdet.py
+++ b/acstools/satdet.py
@@ -128,7 +128,6 @@ jc8m10syq_flc.fits[6] updated
 #
 
 # STDLIB
-from enum import KEEP
 import glob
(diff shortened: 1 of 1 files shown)


jryon (approved):
Looks good to me. I suggest we hold off on making a new release for this fix (unless we start getting help calls) and combine it with future updates/fixes.

PR #201, opened 2024-02-29
update_dq improvement in findsat_mrt

dvstark:
### Description

This PR is to add in functionality to acstools.utils_findsat_mrt.updatedq to expand the input mask to match the size of the original DQ array if necessary. This is needed if a satellite trail was identified and masked on a binned image, which is common. Users could do this manually, but it is needed frequently enough that it made sense to include. This step has also been designed to handle images with odd dimensions,(e.g., WFC3/UVIS are 2051x4096) which ensures the code is applicable to other imaging data. 

Warnings have been added into acstools.findsat_mrt.TrailFinder.rebin step to inform users if their data has odd dimensions and that rows/columns were being trimmed as needed to allow rebinning (this was happening already, but users may not have been aware). These trimmed rows/columns are added back in during the update_dq step after the satellite mask is enlarged.

### DIFF
--- a/acstools/findsat_mrt.py
+++ b/acstools/findsat_mrt.py
@@ -1360,6 +1360,11 @@ class WfcWrapper(TrailFinder):
 
         LOG.info('Rebinning the data by {}'.format(self.binsize))
 
+        # trigger a warning if the image dimensions are not even
+        if (self.image.shape[0] % self.binsize != 0) | (self.image.shape[1] % self.binsize != 0):
+            LOG.warning('Warning: image dimensions do not evenly divide by bin size.\n'
+                        'Extra rows/columns will be trimmed first')
+
         # setting a warning filter for the nansum calculations. These
         # warnings are inconsequential and already accounted for in the code
         with warnings.catch_warnings():
--- a/acstools/findsat_mrt.py
+++ b/acstools/findsat_mrt.py
@@ -1290,8 +1290,8 @@ class WfcWrapper(TrailFinder):
 
     @binsize.setter
     def binsize(self, value):
-        if value is not None and not isinstance(value, int):
-            raise ValueError(f"binsize must be None or int but got: {value}")
+        if value is not None and not (isinstance(value, int) and value >= 1):
+            raise ValueError(f"binsize must be None or int >= 1 but got: {value}")  # noqa
         self._binsize = value
 
     @property
@@ -1361,8 +1361,8 @@ class WfcWrapper(TrailFinder):
         LOG.info('Rebinning the data by {}'.format(self.binsize))
 
         # trigger a warning if the image dimensions are not even
-        if ((self.image.shape[0] % self.binsize != 0)) |
-           ((self.image.shape[1] % self.binsize != 0)):
+        if (((self.image.shape[0] % self.binsize != 0)) |
+           ((self.image.shape[1] % self.binsize != 0))):
             LOG.warning('Image dimensions do not evenly divide by bin size.'
                         '\nExtra rows/columns will be trimmed first.')
 
--- a/acstools/findsat_mrt.py
+++ b/acstools/findsat_mrt.py
@@ -1361,10 +1361,10 @@ class WfcWrapper(TrailFinder):
         LOG.info('Rebinning the data by {}'.format(self.binsize))
 
         # trigger a warning if the image dimensions are not even
-        if ((self.image.shape[0] % self.binsize != 0)) | \
+        if ((self.image.shape[0] % self.binsize != 0)) |
            ((self.image.shape[1] % self.binsize != 0)):
-            LOG.warning('Warning: image dimensions do not evenly divide by bin'
-                        'size.\n Extra rows/columns will be trimmed first')
+            LOG.warning('Image dimensions do not evenly divide by bin size.'
+                        '\nExtra rows/columns will be trimmed first.')
 
         # setting a warning filter for the nansum calculations. These
         # warnings are inconsequential and already accounted for in the code
--- a/acstools/findsat_mrt.py
+++ b/acstools/findsat_mrt.py
@@ -1361,9 +1361,10 @@ class WfcWrapper(TrailFinder):
         LOG.info('Rebinning the data by {}'.format(self.binsize))
 
         # trigger a warning if the image dimensions are not even
-        if (self.image.shape[0] % self.binsize != 0) | (self.image.shape[1] % self.binsize != 0):
-            LOG.warning('Warning: image dimensions do not evenly divide by bin size.\n'
-                        'Extra rows/columns will be trimmed first')
+        if ((self.image.shape[0] % self.binsize != 0)) | \
+           ((self.image.shape[1] % self.binsize != 0)):
+            LOG.warning('Warning: image dimensions do not evenly divide by bin'
+                        'size.\n Extra rows/columns will be trimmed first')
 
         # setting a warning filter for the nansum calculations. These
         # warnings are inconsequential and already accounted for in the code
@@ -1411,4 +1412,4 @@ class WfcWrapper(TrailFinder):
             raise ValueError(f'DQ array can only be updated for FLC/FLT images, not {self.image_type}')  # noqa
 
         update_dq(self.image_file, self.extension + 2, self.mask, dqval=dqval,
-                  verbose=verbose)
+                  verbose=verbose, expand_mask=True)
--- a/acstools/utils_findsat_mrt.py
+++ b/acstools/utils_findsat_mrt.py
@@ -1234,29 +1234,41 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True,
                 return
             else:
                 LOG.info('Enlarging mask to original size.')
+
+                # Determine factor by which to scale-up the image. Scales should
+                # be the same for the x and y dimensions, after we round down
+                # to the nearest integer in case there are extra rows/columns 
+                # that do not divide evenly. These extra rows/columns will be 
+                # added back at the end. This approach is the reverse of
+                # astropy.nddata.block_reduce, which trims extra rows/cols
+                # first. 
+
                 factor_x = np.floor(dqarr.shape[1]/mask.shape[1]).astype(int)
                 factor_y = np.floor(dqarr.shape[0]/mask.shape[0]).astype(int)
-                mask = block_replicate(mask, [factor_y, factor_x],
+                if factor_x != factor_y:
+                    raise ValueError('Mask dimensions require scaling by'
+                                     'different amounts in order to match the DQ'
+                                     'array but this is not allowed.')
+                else:
+                    factor = factor_x
+
+                # assess whether there will be extra rows/columns after scaling
+                extra_cols = dqarr.shape[1] - mask.shape[1]*factor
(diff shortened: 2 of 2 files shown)


pllim:
Lemme close/reopen and see.

pllim:
You can ignore the triage failure in this PR.  That is my bad.

dvstark:
just to put all comments in one space: I had asked what the triage step is and what might be causing it to fail

acstools/findsat_mrt.py
```diff
@@ -1360,6 +1360,12 @@ def rebin(self):
 
         LOG.info('Rebinning the data by {}'.format(self.binsize))
 
+        # trigger a warning if the image dimensions are not even
+        if ((self.image.shape[0] % self.binsize != 0)) | \
+           ((self.image.shape[1] % self.binsize != 0)):
```
pllim, line 1365:
```suggestion
        if (((self.image.shape[0] % self.binsize != 0)) |
               ((self.image.shape[1] % self.binsize != 0))):
```

acstools/findsat_mrt.py
```diff
@@ -1360,6 +1360,12 @@ def rebin(self):
...
+            LOG.warning('Warning: image dimensions do not evenly divide by bin'
+                        'size.\n Extra rows/columns will be trimmed first')
```
pllim, line 1367: No need to put "Warning:" because logger will make that clear.
```suggestion
            LOG.warning('Image dimensions do not evenly divide by bin '
                        'size.\nExtra rows/columns will be trimmed first.')
```

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,54 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
         tailored for ACS/WFC.
     verbose : bool, optional
         Print extra information to the terminal.
+    expand_mask: int, optional
```
pllim, line 1203:
```suggestion
    expand_mask : bool, optional
```

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,54 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
...
+        Allows the mask to be expanded to match the size of the original mask.
+        This is relevant for masks that were generated used binned versions of
+        the original image
```
pllim, line 1206:
```suggestion
        the original image.
```

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,54 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
...
+
+        if not same_size:
+            LOG.info('Inconsistent mask sizes: \n'
+                     'Input mask: {} \n'
+                     'DQ array: {}'.format(mask.shape, dqarr.shape))
+            if not expand_mask:
+                LOG.warning('Cannot proceed due to size mismatch.\n'
+                            'Set expand_mask=True if the mask was generated '
+                            'with binned data and needs to be enlarged to the '
+                            'original size')
```
pllim, line 1223:
```suggestion
                            'original size.')
```

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,54 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
...
+                return
+            else:
+                LOG.info('Enlarging mask to original size')
```
pllim, line 1226:
```suggestion
                LOG.info('Enlarging mask to original size.')
```

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,54 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
...
+                factor_x = np.floor(dqarr.shape[1]/mask.shape[1]).astype(int)
```
pllim, line 1227: The logger saying "enlarging" but I don't see any checks on whether one dimension is really larger than the other one. Won't it be shrinking in some cases? Should that be allowed?

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,54 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
...
+                factor_y = np.floor(dqarr.shape[0]/mask.shape[0]).astype(int)
+                mask = block_replicate(mask, [factor_y, factor_x],
+                                       conserve_sum=False)
+
+                # for original data that had odd dimensions before binning, the
+                # mask still will not match the original. Duplicate end
+                # rows/columns in this case
+
+                same_size = mask.shape == dqarr.shape
```
pllim, line 1236: Instead of doing it twice, is there a smarter way to check for odd/even at the start and then go different logic routes from there?

pllim (commented):
You do have a minimal test in https://github.com/spacetelescope/acstools/blob/master/acstools/tests/test_wfc_findsat_mrt.py . If you can expand your test suite to cover this new option, would prevent regression in the future, but I'll leave that up to you.

acstools/utils_findsat_mrt.py
jryon: Seconding this. I did a couple of tests and the error for a mask that is bigger than the DQ array is pretty cryptic. If you never expect this case to actually happen, should handle it more elegantly.

jryon (changes requested):
I ran a few basic tests with different sizes of masks paired with ACS and WFC3 data. The results look good and make sense, except for the case when the mask is larger than the DQ array.

acstools/utils_findsat_mrt.py
dvstark: Thanks y'all. I can't think of a situation where someone would want to supersample the original image to make it larger for satellite trail detection (feel free to chime in if you think I'm thinking too narrowly). I think the best path forward is to enforce self.binsize > 1, which will always yield a mask smaller or equal in size to the DQ array. I can add some checks inside the updatedq() function to check this as well.

acstools/utils_findsat_mrt.py
dvstark: I cleaned this up a bit, but it's more like an extra step if the images don't divide evenly, rather than completely different logic

dvstark:
I'm unsure what to do about the error in the "Python 3.12 with remote data and dev dependencies" CI test:py312-test-

```
devdeps: commands[3] /home/runner/work/acstools/acstools/.tmp/py312-test-devdeps> pytest --pyargs acstools /home/runner/work/acstools/acstools/doc --remote-data -v
ImportError while loading conftest '/home/runner/work/acstools/acstools/conftest.py'.
../../conftest.py:10: in <module>
    from acstools.version import version
../../acstools/__init__.py:20: in <module>
    from . import satdet  # noqa
../../acstools/satdet.py:152: in <module>
    import skimage
../../.tox/py312-test-devdeps/lib/python3.12/site-packages/skimage/__init__.py:128: in <module>
    from ._shared import geometry
geometry.pyx:1: in init skimage._shared.geometry
    ???
E   ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
py312-test-devdeps: exit 4 (1.01 seconds) /home/runner/work/acstools/acstools/.tmp/py312-test-devdeps> pytest --pyargs acstools /home/runner/work/acstools/acstools/doc --remote-data -v pid=2518
  py312-test-devdeps: FAIL code 4 (303.45=setup[272.30]+cmd[9.07,20.65,0.42,1.01] seconds)
  evaluation failed :( (303.59 seconds)
Error: Process completed with exit code 4.
```

Some new issue with skimage and satdet?

pllim:
You can ignore the devdeps failure for now. Maybe scikit-image nightly wheel needs time to catch up to numpy nightly wheel. With numpy 2.0 in beta now, hopefully soon. I'll give it a few days and up to a week.

dvstark:
I've done the updates and am ready for this PR to be reviewed again. Thanks for the feedback so far!

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,75 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
...
+        # provide informative error messages depending on mask/dqarr size
+        if not same_size:
+            LOG.info('Inconsistent mask sizes: \n'
+                     'Input mask: {} \n'
+                     'DQ array: {}'.format(mask.shape, dqarr.shape))
+
+            if mask_larger:
+                LOG.error('Mask is larger than the DQ array in one or more '
+                          'dimensions.\nThis routine cannot currently rescale'
+                          'masks that are larger than the origin input image')
```
jryon, line 1226: need a space between 'rescale' and 'masks'

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,75 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
...
+                # added back at the end. This approach is the reverse of
+                # astropy.nddata.block_reduce, which trims extra rows/cols
+                # first.
+
+                factor_x = np.floor(dqarr.shape[1]/mask.shape[1]).astype(int)
+                factor_y = np.floor(dqarr.shape[0]/mask.shape[0]).astype(int)
+                if factor_x != factor_y:
+                    raise ValueError('Mask dimensions require scaling by'
+                                     'different amounts in order to match the'
+                                     'DQ array but this is not allowed.')
```
jryon, line 1251: need spaces between 'by' and 'different, 'the' and 'DQ'

jryon: I tried a square mask (1000x1000), and got this error. I think the error could be more helpful, but I'm not sure how. Perhaps something to do with the aspect ratio of the mask must be close to that of the DQ array?

jryon (commented):
Couple more very minor comments

acstools/utils_findsat_mrt.py
dvstark: thanks Jenna. Do you meant the error message itself could be more helpful? It could say something like "Mask aspect ratio too different from DQ array" for instance.

jryon: Yes, that's what I mean. Essentially, the message should help the user figure out how to fix the problem. Could think about printing out the aspect ratio of the mask and the dq array, and tell the user to make them more similar?

dvstark: Ok, I modified that error message. Let me know what you think

jryon: Looks good to me!

jryon (approved):
Happy with this!

dvstark:
@pllim The Wheel-building step is being skipped. Do we need to make sure that passes before merging this in?

pllim:
Does not hurt. I added the label to enable it.

acstools/utils_findsat_mrt.py
```diff
@@ -1199,10 +1200,85 @@ def update_dq(filename, ext, mask, dqval=16384, verbose=True):
         tailored for ACS/WFC.
     verbose : bool, optional
         Print extra information to the terminal.
+    expand_mask: bool, optional
```
pllim, line 1203:
```suggestion
    expand_mask : bool, optional
```

pllim (commented):
Since Jenna approved, I have nothing much to add, except one Sphinx formatting fix.

PR #209, opened 2024-08-10
TST: Clean up devdeps, use photutils nightly wheel

pllim:
xref https://github.com/astropy/photutils/pull/1834 

And also clean up devdeps setup.


pllim:
Seems to work. Merging.

PR #210, opened 2024-08-27
replace interp2d in utils_findsat_mrt

dvstark:
### Description

This pull request replaces the old interpolation method in acstools.utils_findsat_mrt.create_mrt_line_kernel from scipy.interpolate.interp2d to scipy.interpolate.RegularGridInterpolator. This request was made by Alyssa Guzman on behalf of DMD because interp2d has been deprecated and they want all packages to be compatible with the latest versions of 3rd party packages. For this update, I used scipy 1.14.1. I have tested the new interpolation method and it returns the same output as the prior method, with some slight variations, but these are very small and acceptable.

### DIFF
--- a/acstools/utils_findsat_mrt.py
+++ b/acstools/utils_findsat_mrt.py
@@ -26,7 +26,6 @@ except ImportError as e:
 # check for scipy
 try:
     from scipy import interpolate
-    from scipy.ndimage import map_coordinates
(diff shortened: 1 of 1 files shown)


pllim (approved):
Technically, LGTM. Thanks!

jryon (approved):
Seems fine to me

PR #258, opened 2026-06-24
Use Trusted Publishing for release

pllim:
### Description

This pull request is to fix #256


.github/workflows/publish-to-pypi.yml
```diff
@@ -47,18 +58,11 @@ jobs:
...
-      run: ls -l dist
+      run: ls -lh dist
 
     - name: Check dist
       run: python -m twine check --strict dist/*
 
-    # FOR DEBUGGING ONLY: repository_url (TestPyPI) and verbose;
-    # Use appropriate token if debugging with TestPyPI
     - name: Publish distribution 📦 to PyPI
       if: (startsWith(github.ref, 'refs/tags/') && github.event_name == 'push')
```
zacharyburnett: this will attempt to upload to pypi on ANY tag, which may fail if the tag is not named something PyPI likes

